PR #10, opened 2015-09-30
Created RestFilter and ActionFilter to update requests to the .kibana…

ewolinetz:
… index to create user profiles and responses to map back to .kibana

### DIFF
--- a/src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexAction.java
+++ b/src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexAction.java
@@ -1,0 +1,116 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.fabric8.elasticsearch.plugin;
+
+import org.elasticsearch.action.ActionListener;
+import org.elasticsearch.action.ActionRequest;
+import org.elasticsearch.action.ActionResponse;
+import org.elasticsearch.action.delete.DeleteResponse;
+import org.elasticsearch.action.get.GetResponse;
+import org.elasticsearch.action.index.IndexResponse;
+import org.elasticsearch.action.support.ActionFilter;
+import org.elasticsearch.action.support.ActionFilterChain;
+import org.elasticsearch.client.Client;
+import org.elasticsearch.cluster.ClusterService;
+import org.elasticsearch.common.inject.Inject;
+import org.elasticsearch.common.logging.ESLogger;
+import org.elasticsearch.common.logging.Loggers;
+import org.elasticsearch.common.settings.Settings;
+import org.elasticsearch.index.get.GetResult;
+
+public class KibanaUserReindexAction implements ActionFilter {
+
+	private final ESLogger logger;
+	private final String kibanaIndex;
+	
+	@Inject
+	public KibanaUserReindexAction(final Settings settings, final ClusterService clusterService, final Client client) {
+		this.logger = Loggers.getLogger(KibanaUserReindexAction.class);
+		this.kibanaIndex = settings.get(ConfigurationSettings.KIBANA_CONFIG_INDEX_NAME, 
+				ConfigurationSettings.KIBANA_CONFIG_INDEX_NAME);
+		
+		logger.debug("Initializing KibanaUserReindexAction");
+	}
+
+	@Override
+	public int order() {
+		// We want this to be the last in the chain
+		return Integer.MAX_VALUE;
+	}
+
+	@Override
+	public void apply(String action, ActionRequest request,
+			ActionListener listener, ActionFilterChain chain) {
+		chain.proceed(action, request, listener);
+	}
+
+	@Override
+	public void apply(String action, ActionResponse response,
+			ActionListener listener, ActionFilterChain chain) {
+		
+		logger.debug("Response with Action '{}' and class '{}'", action, response.getClass());
+		
+		if ( response instanceof IndexResponse ) {
+			final IndexResponse ir = (IndexResponse) response;
+			String index = getIndex(ir);
+			
+			response = new IndexResponse(index, ir.getType(), ir.getId(), ir.getVersion(), ir.isCreated());
+		}
+		else if ( response instanceof GetResponse ) {
+			response = new GetResponse(buildNewResult((GetResponse) response));
+		}
+		else if ( response instanceof DeleteResponse ) {
+			final DeleteResponse dr = (DeleteResponse) response;
+			String index = getIndex(dr);
+			
+			response = new DeleteResponse(index, dr.getType(), dr.getId(), dr.getVersion(), dr.isFound());
+		}
+		
+		chain.proceed(action, response, listener);
+	}
+	
+	private GetResult buildNewResult(GetResponse response) {
+		String index = getIndex(response);
+		
+		GetResult getResult = new GetResult(index, response.getType(), response.getId(), response.getVersion(),
+				response.isExists(), 
+				(!response.isSourceEmpty()) ? response.getSourceAsBytesRef() : null, 
+				response.getFields());
+		
+		return getResult;
+	}
+	
+	private boolean isKibanaUserIndex(String index) {
+		return (index.startsWith(kibanaIndex) && !index.equalsIgnoreCase(kibanaIndex));
+	}
+	
+	private String getIndex(ActionResponse response) {
+		String index = "";
+		
+		if ( response instanceof IndexResponse )
+			index = ((IndexResponse) response).getIndex();
+		else if ( response instanceof GetResponse )
+			index = ((GetResponse)response).getIndex();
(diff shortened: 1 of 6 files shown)


jimmidyson:
add to whitelist

src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexAction.java
```diff
@@ -0,0 +1,103 @@
...
+import org.elasticsearch.common.settings.Settings;
+import org.elasticsearch.index.get.GetResult;
+
+public class KibanaUserReindexAction implements ActionFilter {
+
+	private final ESLogger logger;
+	
+	@Inject
+	public KibanaUserReindexAction(final Settings settings, final ClusterService clusterService, final Client client) {
+		logger = Loggers.getLogger(KibanaUserReindexAction.class, settings);
```
jcantrill: What does the settings add to getting the logger? should we be doing this everywhere?

src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexAction.java
```diff
@@ -0,0 +1,103 @@
...
+	@Override
+	public void apply(String action, ActionResponse response,
+			ActionListener listener, ActionFilterChain chain) {
+		
+		logger.debug("Response with Action '{}' and class '{}'", action, response.getClass());
+		
+		if ( response instanceof IndexResponse ) {
+			String index = ((IndexResponse) response).getIndex();
+			if ( isKibanaUserIndex(index) ) {
+				index = ".kibana";
```
jcantrill: We should allow this to be configurable since its a configurable setting within a Kibana installation

src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexAction.java
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	@Override
+	public void apply(String action, ActionResponse response,
+			ActionListener listener, ActionFilterChain chain) {
+		
+		logger.debug("Response with Action '{}' and class '{}'", action, response.getClass());
+		
+		if ( response instanceof IndexResponse ) {
+			String index = ((IndexResponse) response).getIndex();
```
jcantrill: Cast once to make more readable since you are using it in logic following this

src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexAction.java
```diff
@@ -0,0 +1,103 @@
...
+				index = ".kibana";
+			}
+			
+			response = new IndexResponse(index, ((IndexResponse) response).getType(), ((IndexResponse) response).getId(), ((IndexResponse) response).getVersion(), ((IndexResponse) response).isCreated());
+		}
+		else if ( response instanceof GetResponse ) {
+			response = new GetResponse(buildNewResult((GetResponse) response));
+		}
+		else if ( response instanceof DeleteResponse ) {
+			String index = ((DeleteResponse) response).getIndex();
```
jcantrill: Cast once.  Same suggestion as above

jcantrill: The logic for getting the index here and above looks like you could roll it into a single method that takes a Generic param like:  String getIndex(T response)

src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexFilter.java
```diff
@@ -0,0 +1,97 @@
...
+import org.elasticsearch.http.netty.NettyHttpRequest;
+import org.elasticsearch.rest.RestChannel;
+import org.elasticsearch.rest.RestFilter;
+import org.elasticsearch.rest.RestFilterChain;
+import org.elasticsearch.rest.RestRequest;
+
+public class KibanaUserReindexFilter extends RestFilter {
+
+	private static final String DEFAULT_AUTH_PROXY_HEADER = "X-Proxy-Remote-User";
+	private static final String SEARCHGUARD_AUTHENTICATION_PROXY_HEADER = "searchguard.authentication.proxy.header";
```
jcantrill: If these are not already in the ConfigurationSettings interface of our plugin, they should be

src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexFilter.java
```diff
@@ -0,0 +1,97 @@
...
+
+		} catch (Exception e) {
+			logger.error("Error handling request in OpenShift SearchGuard filter", e);
+		} finally {
+			chain.continueProcessing(request, channel);
+		}
+	}
+
+	private String getUser(RestRequest request) {
+		return (String) ObjectUtils.defaultIfNull(request.header(proxyUserHeader), "");
```
jcantrill: There is an equivalent method in StringUtils that would not require you to cast

src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexFilter.java
```diff
@@ -0,0 +1,97 @@
...
+	
+	@Override
+	public void process(RestRequest request, RestChannel channel, RestFilterChain chain) throws Exception {
+		try {
+			logger.debug("Handling Request in Kibana User Reindex filter...");
+			final String user = getUser(request);
+			final String requestedIndex = getRequestedIndex(request);
+			
+			// rewrite .kibana indexes to .kibana.<username> based on the user header 
+			if ( StringUtils.isNotEmpty(user) && StringUtils.isNotEmpty(requestedIndex) && requestedIndex.equalsIgnoreCase(".kibana")) {
```
jcantrill: Kibana should be configurable (prefered) or at a minimum made into a const

src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexFilter.java
```diff
@@ -0,0 +1,97 @@
...
+		
+		NettyHttpRequest nettyRequest = (NettyHttpRequest) request;
+		HttpRequest httpRequest = nettyRequest.request();
+		httpRequest.setUri(replaced);
+		
+		return new NettyHttpRequest(httpRequest, nettyRequest.getChannel());
+	}
+	
+	private String getRequestedIndex(RestRequest request) {
+		String uri = (String) ObjectUtils.defaultIfNull(request.uri(), "");
```
jcantrill: StringUtils

src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexFilter.java
```diff
@@ -0,0 +1,97 @@
...
+		String replaced = uri.replace(oldIndex, newIndex);
+		
+		NettyHttpRequest nettyRequest = (NettyHttpRequest) request;
+		HttpRequest httpRequest = nettyRequest.request();
+		httpRequest.setUri(replaced);
+		
+		return new NettyHttpRequest(httpRequest, nettyRequest.getChannel());
+	}
+	
+	private String getRequestedIndex(RestRequest request) {
```
jcantrill: Consider looking at java.net.uri.  I'm wondering if you can do the same thing using something from the class library that might be safer then what we are doing here.

src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexAction.java
ewolinetz: I had copied it from the strategy used to create the logger in OpenShiftElasticSearchService.java, it may not be necessary.  Will confirm if it can be removed.

src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexFilter.java
ewolinetz: this is a little misleading I think, request.uri is just a string, and if we used the java.net.uri package we would end up creating a new URI object and parsing it to get the path which is what we already get from request.uri...

src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexFilter.java
```diff
@@ -0,0 +1,96 @@
...
+		}
+	}
+
+	private String getUser(RestRequest request) {
+		return StringUtils.defaultIfBlank(request.header(proxyUserHeader), "");
+	}
+	
+	private RestRequest updateRequestIndex(RestRequest request, String oldIndex, String newIndex) {
+		String uri = request.uri();
+		String replaced = uri.replace(oldIndex, newIndex);
```
sosiouxme: Can't find the java docs on this... does this replace the value in a URI-safe way? In case the user name has characters that need escaping.

ewolinetz: The api docs [1] state that it replaces the literal target with the literal replacement sequence.  So it shouldn't need to do any character escaping...  

[1] http://docs.oracle.com/javase/7/docs/api/java/lang/String.html#replace%28java.lang.CharSequence,%20java.lang.CharSequence%29

jcantrill:
LGTM to merge after you have resolved @sosiouxme question

src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexFilter.java
sosiouxme: Somehow I missed that it was just a String. Thought it was a URI method or something.

OK, in that case we need to escape newIndex before putting it in the URI string. Because it could be `.kibana.doesItBlend?yesItDoes` and that needs to not choke as a URI.

ewolinetz: We now pass in the newIndex as ".kibana." + the sha1 hash of the user name since it may contain characters that ES doesn't allow in an index. (line 51)

ewolinetz:
[merge]

PR #15, opened 2015-10-13
Evaluate if a user should see operations index

jcantrill:
This PR adds the evaluation of a users ability to see the operations logs based on their role of cluster-admin.

@sosiouxme @ewolinetz @jimmidyson  please review.

Note: once the fabric8 client supports SAR, the isClusterAdmin can be rewritten.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,5 @@ crashlytics-build.properties
 
 
 /test/
+/bin/
+/bin/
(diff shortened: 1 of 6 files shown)


ewolinetz:
@jcantrill fyi unless you think we'll need it for something else, I don't need to pull roles out of context.  My kibana seeding takes place in this same block of code, so I can just use isClusterAdmin for the time being (until the SAR is available).  Other than that LGTM.

jcantrill:
@ewolinetz It would be beneficial to use the context version if possible so we don't make a second call to the API server to get the same information.

ewolinetz:
@jcantrill +1  [merge]

PR #16, opened 2015-10-22
Fix kibana seeding

ewolinetz:
@jcantrill @sosiouxme @jimmidyson PTAL.
We now correctly seed the kibana index for a user based on their OS project access.
We also remove a user and reset their default index if their previous default was removed due to access being removed on OS.

FYI: all but the latest 3 commits are throw away :)

### DIFF
--- a/src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
+++ b/src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
@@ -15,6 +15,7 @@
  */
 package io.fabric8.elasticsearch.plugin.acl;
 
+import static io.fabric8.elasticsearch.plugin.kibana.KibanaSeed.setDashboards;
 import io.fabric8.elasticsearch.plugin.ConfigurationSettings;
 import io.fabric8.kubernetes.client.ConfigBuilder;
 import io.fabric8.openshift.api.model.Project;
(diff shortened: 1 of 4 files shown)


src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
```diff
@@ -0,0 +1,286 @@
...
+			// cull any that are in ES but not in OS (remaining in indexPatterns)
+			remove(user, indexPatterns, esClient, kibanaIndex);
+			
+			Collections.sort(common);
+			// Set default index to first index in common if we removed the default
+			String defaultIndex = getDefaultIndex(user, esClient, kibanaIndex, kibanaVersion);
+			if ( indexPatterns.contains(defaultIndex) || StringUtils.isEmpty(defaultIndex) ) {
+				for ( String project : common ) {
+					setDefaultIndex(user, project, esClient, kibanaIndex, kibanaVersion);
+					break;
```
jcantrill: i dont think this is needed. means you would only process 1 project

ewolinetz: common used to be a set and i didn't' update the logic.. will fix

src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
```diff
@@ -128,6 +129,8 @@ public void process(RestRequest request, RestChannel channel, RestFilterChain ch
 				if(updateCache(user, token, isClusterAdmin)){
 					syncAcl();
 				}
+
+				setDashboards(user, listProjectsFor(token), roles, esClient, kibanaIndex, kibanaVersion);
```
ewolinetz: move this so we don't have to make a call to get projects twice; we shouldn't be updating the dashboards unless we update the cache anyways

ewolinetz:
@jcantrill addressed nits

jcantrill:
LGTM [merge]

ewolinetz:
[remerge]

ewolinetz:
rebased [merge]

PR #17, opened 2015-10-27
Making changes so that plugin will seed initial ACL config instead of…

ewolinetz:
… using run.sh method

@jcantrill @sosiouxme @jimmidyson  PTAL


jcantrill:
LGTM

sosiouxme:
Seems to work in clustered setup.

PR #18, opened 2015-11-03
Adding a default index of .empty-project; fixing how we check is a ne…

ewolinetz:
…w default index is required

@jcantrill @jimmidyson PTAL


jcantrill:
LGTM [merge]

Issue #20, opened 2016-02-15
Question regarding createSearchProperties


danbev:
We are using this plugin as port of [origin-aggregated-logging](https://github.com/openshift/origin-aggregated-logging/blob/master/elasticsearch/install.sh#L18) and hitting an issue which I think is related to [createSearchProperties](https://github.com/fabric8io/openshift-elasticsearch-plugin/blob/master/src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java#L92) and the [content](https://github.com/fabric8io/openshift-elasticsearch-plugin/blob/master/src/main/java/io/fabric8/elasticsearch/plugin/kibana/DocumentBuilder.java#L23) it adds to the ```.kibana/search``` index. 
What we are seeing that it is not possible to save searches in Kibana. Trying to do so gives the following error message:
```shell
Discover: Request to Elasticsearch failed: "MapperParsingException[object mapping for [search] tried to parse field [title] as object, but got EOF, has a concrete value been provided to it?]"

Error: Request to Elasticsearch failed: "MapperParsingException[object mapping for [search] tried to parse field [title] as object, but got EOF, has a concrete value been provided to it?]"
    at https://kibana.local.feedhenry.io/index.js?_b=7562:88457:15
    at wrappedErrback (https://kibana.local.feedhenry.io/index.js?_b=7562:20902:78)
    at wrappedErrback (https://kibana.local.feedhenry.io/index.js?_b=7562:20902:78)
    at https://kibana.local.feedhenry.io/index.js?_b=7562:21035:76
    at Scope.$eval (https://kibana.local.feedhenry.io/index.js?_b=7562:22022:28)
    at Scope.$digest (https://kibana.local.feedhenry.io/index.js?_b=7562:21834:31)
    at Scope.$apply (https://kibana.local.feedhenry.io/index.js?_b=7562:22126:24)
    at done (https://kibana.local.feedhenry.io/index.js?_b=7562:17661:45)
    at completeRequest (https://kibana.local.feedhenry.io/index.js?_b=7562:17875:7)
    at XMLHttpRequest.xhr.onreadystatechange (https://kibana.local.feedhenry.io/index.js?_b=7562:17814:11)
```
After enabling logging we can see the following in the ElasticSearch log:
```shell
[...]

PR #21, opened 2016-02-17
Removing the indexing of .kibana/search/properties

danbev:
Motivation:
When trying to save a search in Kibana the following error is displayed:

Discover: Request to Elasticsearch failed:
"MapperParsingException[object mapping for [search] tried to parse field
[title] as object, but got EOF, has a concrete value been provided to
it?]"

The .kibana index has a type named 'search' which it uses to store
searches made in the Discover view of the Kibana Web Console.

When createSearchProperties is called it inserts a document into
.kibana/search with an id of 'properties'. This data is then inspected by
ElasticSearch and it will dynamically create a mapping for those fields.

Looking at the contents of the sent data it looks more like type mapping
which one might expect to see in .kibana/_mappings/search, and it does
not look like format of a search saved by Kibana. The field names are the same in
this case but instead of expecting the 'title' field to be of type
string, it is expecting it to be of type object as that is what the
mapping was updated to.

Modifications:
Removed the createSearchProperties call from the code.

Result:
After this it is possible to save searches in the Kibana console.

Issue:
https://github.com/fabric8io/openshift-elasticsearch-plugin/issues/20


jcantrill:
@ewolinetz Are you able to validate this is an acceptable modification? @danbev I do not recall if there are unit tests associated with this logic.

danbev:
@jcantrill I had a look but was unable to find any in the project itself.

ewolinetz:
Closing this PR as it was addressed as part of #22

PR #22, opened 2016-02-17
Tech-debt for SAR and address #20

ewolinetz:


jcantrill:
@ewolinetz LGTM.  We I would leave as two commits in case we only want to cherry-pick the seeding change into the maintenance release.

ewolinetz:
@jcantrill +1 I was planning on leaving it as two separate commits since they're not related features.  I don't think we'll need to do a maintenance release though

ewolinetz:
[merge]

PR #26, opened 2016-04-13
Adding config properties to disable the dynamic ACL, Kibana seeding, …

ewolinetz:
…and Kibana user rewrites

### DIFF
--- a/src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexAction.java
+++ b/src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexAction.java
@@ -54,12 +54,16 @@ public class KibanaUserReindexAction implements ActionFilter, ConfigurationSetti
 	private final ESLogger logger;
 	private final String kibanaIndex;
 	
+	private Boolean enabled;
+	
 	@Inject
 	public KibanaUserReindexAction(final Settings settings, final ClusterService clusterService, final Client client) {
+		this.enabled = settings.getAsBoolean(OPENSHIFT_KIBANA_REWRITE_ENABLED_FLAG, OPENSHIFT_KIBANA_REWRITE_ENABLED_DEFAULT);
 		this.logger = Loggers.getLogger(KibanaUserReindexAction.class);
 		this.kibanaIndex = settings.get(KIBANA_CONFIG_INDEX_NAME, DEFAULT_USER_PROFILE_PREFIX);
 		
-		logger.debug("Initializing KibanaUserReindexAction");
+		if ( enabled )
+			logger.debug("Initializing KibanaUserReindexAction");
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


ewolinetz:
@sosiouxme @jimmidyson PTAL

src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexAction.java
```diff
@@ -78,69 +82,71 @@ public void apply(String action, ActionRequest request,
...
-			else if ( response instanceof MultiGetResponse ) {
-				final MultiGetResponse mgr = (MultiGetResponse) response;
-				
-				MultiGetItemResponse[] responses = new MultiGetItemResponse[mgr.getResponses().length];
-				int index = 0;
-				
-				for ( MultiGetItemResponse item : mgr.getResponses() ) {
-					GetResponse itemResponse = item.getResponse();
-					Failure itemFailure = item.getFailure();
+		if ( enabled ) {
```
jcantrill: Is this a safety fallback?  I see we are skipping registration in the service class if the configuration is FALSE.

src/main/java/io/fabric8/elasticsearch/plugin/OpenShiftElasticSearchService.java
```diff
@@ -65,31 +66,41 @@ protected OpenShiftElasticSearchService(final Settings settings, final Client es
...
+		if ( reindexEnabled ) {
+			restController.registerFilter(new KibanaUserReindexFilter(settings, logger));
+		}
 	}
 
 	@Override
 	protected void doStart() throws ElasticsearchException {
+		
+		boolean dynamicEnabled = settings.getAsBoolean(OPENSHIFT_DYNAMIC_ENABLED_FLAG, OPENSHIFT_DYNAMIC_ENABLED_DEFAULT);
+
```
jcantrill: Are you sure we need the notifier thread?  Originally, this was provided as a way to provide inter-thread communication where one was blocked awaiting for a 'callback' from the other.  I would think we would not need it if we are not rewriting the ACL

jcantrill: We additionally should provide debug log message with what the configuration was interpreted to be.  Might make it more obvious that are enabling (or not) the functionality

src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexAction.java
ewolinetz: We're doing this here because the module where we add this Action doesn't have the settings injected so unfortunately we can't prevent this from being added then.

src/main/java/io/fabric8/elasticsearch/plugin/OpenShiftElasticSearchService.java
ewolinetz: We are still letting users seed their initial ACL even if they disable the dynamic ACL.  From what I tried there wasn't a way to trigger that logic with the notifier disabled too.

I can add a debug statement if features are disabled/enabled.

src/main/java/io/fabric8/elasticsearch/plugin/OpenShiftElasticSearchService.java
```diff
@@ -65,31 +66,53 @@ protected OpenShiftElasticSearchService(final Settings settings, final Client es
...
+			logger.debug("Starting with Kibana reindexing feature disabled");
+		}
 	}
 
 	@Override
 	protected void doStart() throws ElasticsearchException {
+		
+		boolean dynamicEnabled = settings.getAsBoolean(OPENSHIFT_DYNAMIC_ENABLED_FLAG, OPENSHIFT_DYNAMIC_ENABLED_DEFAULT);
+
+		if ( dynamicEnabled ) {
```
jcantrill: how about just 
`logger.debug(String.format("Starting with Dynamic ACL feature enabled: %s", dynamicEnabled))`

ewolinetz: That is more simple, I can do that... i was wanting it to explicitly say 'enabled' or 'disabled'

ewolinetz:
@jcantrill Addressed logging comments

PR #34, opened 2016-08-25
ensuring to use empty scope for checking if user is cluster-admin

ewolinetz:
@jcantrill this is the es 2.3 version of #33


ewolinetz:
not as high of a priority

jcantrill:
LGTM [merge]

PR #35, opened 2016-09-01
removing all alias if user is not cluster-admin

ewolinetz:
@jcantrill ptal

`indexPatterns` list gets passed along to delete patterns from the contents


jcantrill:
[merge]  A test would be nice..but in the interest of time...

ewolinetz:
@jcantrill ack -- will take on as tech-debt

PR #38, opened 2016-09-06
Adding null check for default index for case that we have changed kib…

ewolinetz:
…ana versions

@jcantrill PTAL


jcantrill:
lgtm [merge]

PR #44, opened 2016-10-05
support the common data model

richm:
This adds 3 new elasticsearch.yml openshift.config section parameters:

* `use_common_data_model` - boolean - default `false` - if true, tell the
  plugin that you want to use the common data model
* `project_index_prefix` - string - default `""` - set this to the string value
  that project/namespace indices use as their prefix - for example, with the
  common data model, if the namespace is `test`, the index name will be
  `project.test.$uuid.YYYY.MM.DD`.  In this case, use `"project."` as the
  prefix.
* `time_field_name` - string - default `"time"` - the common data model uses
  `"@timestamp"` as the name of the time field

If the common data model is used, the common data model index-pattern values
will be loaded to match the common data model index naming, model, and field
naming conventions.

The index-mapping values are from doing a search of the .kibana index after
loading Kibana and loading data from common data model indices.


richm:
@ewolinetz @jcantrill PTAL

richm:
@ewolinetz In addition to these changes, are there any acl/search guard related changes needed?

ewolinetz:
@richm yes, it looks like the logic for that is missing from here -- and it may change the seeding logic slightly. The ACL is defined and updated within the dynamic_acl_filter

richm:
@ewolinetz are you talking about this code from `public void syncFrom`?
```
	public void syncFrom(UserProjectCache cache, final String userProfilePrefix){
...
			String indexName = String.format("%s?*", project.replace('.', '?'));
...
			String indexName = String.format("%s?%s", userProfilePrefix.replace('.', '?'), usernameHash);
```
Those index names need to have the `project.` prefix?

richm:
yeah, not sure what's going on.  I created a testuser that has admin rights in the test project.  If I login to the openshift console, I can only see the test project.  However, in kibana/es, I can view project.logging.* and even .operations.*

richm:
I went back to 1.3 so I could see how access control is supposed to work, but I'm having a hard time testing access control with 1.3.  This is what I did:
```
oc new-project test
oc new-app centos/ruby-22-centos7~https://github.com/openshift/ruby-ex.git
oc login --username=testuser --password=testuser
test_token="$(oc whoami -t)"
test_name="$(oc whoami)"
test_ip="127.0.0.1"
oc login --username=system:admin
oc policy add-role-to-user admin testuser -n test
oc exec -n logging logging-kibana-1-napr6 -c kibana -- curl -s -k --cert /etc/kibana/keys/cert --key /etc/kibana/keys/key -H "X-Proxy-Remote-User: $test_name" -H "Authorization: Bearer $test_token" -H "X-Forwarded-For: 127.0.0.1" https://logging-es:9200/test.*/_search | python -mjson.tool
{
    "error": "UnknownHostException[No trusted proxies]",
    "status": 500
}
```
but if I omit the proxy stuff, it works:
```
oc exec -n logging logging-kibana-1-napr6 -c kibana -- curl -s -k        --cert /etc/kibana/keys/cert --key /etc/kibana/keys/key        https://logging-es:9200/test.*/_search | python -mjson.tool
{
    "_shards": {
        "failed": 0,
        "successful": 5,
        "total": 5
    },
    "hits": {
        "hits": [
            {
...
            }
        ],
        "max_score": 1.0,
        "total": 45
```

src/main/java/io/fabric8/elasticsearch/plugin/kibana/DocumentBuilder.java
```diff
@@ -54,14 +57,20 @@ public DocumentBuilder intervalName(String name) {
 		return addField("intervalName", name);
 	}
 	
-	public DocumentBuilder applicationFields() {
-		return addField("fields", StringEscapeUtils.escapeJava(APPLICATION_FIELDS));
+	public DocumentBuilder applicationFields(boolean use_cdm) {
+		if ( false == use_cdm )
```
jcantrill: This is an odd coding convention for java though I like the explicitness of it.  Consider using !use_cdm but its not a deal breaker for me.

src/main/java/io/fabric8/elasticsearch/plugin/kibana/DocumentBuilder.java
```diff
@@ -54,14 +57,20 @@ public DocumentBuilder intervalName(String name) {
...
+			return addField("fields", StringEscapeUtils.escapeJava(APPLICATION_FIELDS));
+		else
+			return addField("fields", StringEscapeUtils.escapeJava(APPLICATION_CDM_FIELDS));
```
jcantrill: We should pull this into a constant since its used in many places.

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
```diff
@@ -48,11 +49,12 @@
 import org.elasticsearch.client.Client;
 import org.elasticsearch.common.logging.ESLogger;
 import org.elasticsearch.common.logging.Loggers;
+import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.index.IndexNotFoundException;
 import org.elasticsearch.search.SearchHit;
 import org.elasticsearch.transport.RemoteTransportException;
 
-public class KibanaSeed {
+public class KibanaSeed implements ConfigurationSettings {
```
jcantrill: Why do we need to impl ConfigurationSettings?  I didnt see any obvious uses of it.

src/main/java/io/fabric8/elasticsearch/plugin/kibana/DocumentBuilder.java
richm: @jcantrill sorry - pull what into a constant?

ewolinetz:
@richm the project we get in that snippet should already have the prefix at that point in time. That is just for formatting the ACL to match searchguard's required format (they don't allow for '.' so we replace them with '?').

If a normal user is able to see everything, then something is wrong with their ACL... if you change the logger level for the kibana seed class, you should see which projects come back for the user.

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
richm: @jcantrill that seems to be the technique that other classes use in order to have access to the constants defined there such as OPENSHIFT_CONFIG_USE_COMMON_DATA_MODEL

ewolinetz: Correct

src/main/java/io/fabric8/elasticsearch/plugin/kibana/DocumentBuilder.java
jcantrill: "fields"

richm:
With this patch plus https://github.com/fabric8io/openshift-elasticsearch-plugin/pull/47 everything is working - testuser can see `project.test.*` but not `project.logging.*` or `.operations.*`:
```
{"error":{"root_cause":[{"type":"security_exception","reason":"no permissions for indices:data/read/search"}],"type":"security_exception","reason":"no permissions for indices:data/read/search"},"status":403}
```
I think I addressed all of the review comments.  Please review.

ewolinetz:
@richm I wouldn't mind doing a release containing just #47 prior to merging this in just in case (call me paranoid).

Otherwise LGTM

richm:
@jcantrill ptal

@ewolinetz ok - yes, let's release #47 asap

src/main/java/io/fabric8/elasticsearch/plugin/acl/SearchGuardRoles.java
```diff
@@ -156,17 +156,23 @@ public void removeRole(Roles role) {
...
 		for (String project : cache.getAllProjects() ) {
 			String projectName = String.format("%s_%s", PROJECT_PREFIX, project.replace('.', '_')); 
 			String indexName = String.format("%s?*", project.replace('.', '?'));
-			
 			RoleBuilder role = new RoleBuilder(projectName)
 					.setActions(indexName, DEFAULT_ROLE_TYPE, DEFAULT_ROLE_ACTIONS);
+
+			// If using common data model, allow access to both the $projname.$uuid.* indices and
+			// the project.$projname.$uuid.* indices for backwards compatability
+			if ( cdm_project_prefix.length() > 0 ) {
```
jcantrill: StringUtils.IsBlank .... or isEmpty... will handle the case of empty string

jcantrill:
Other then minor nit... LGTM

richm:
I found a couple of other minor issues.  @ewolinetz @jcantrill ptal

jcantrill:
LGTM

README.md
```diff
@@ -43,6 +43,24 @@ openshift.operations.project.names: ["default", "openshift", "openshift-infra"]
...
+If your OpenShift EFK cluster supports the common data model, and you want to
+use the common data model, there are some configuration parameters that you can
+set in `elasticsearch.yml`, under `openshift.config`:
+
+* `use_common_data_model` - boolean - default `false` - if true, tell the
+  plugin that you want to use the common data model
+* `project_index_prefix` - string - default `""` - set this to the string value
+  that project/namespace indices use as their prefix - for example, with the
+  common data model, if the namespace is `test`, the index name will be
+  `project.test.$uuid.YYYY.MM.DD`.  In this case, use `"project."` as the
```
ewolinetz: technically shouldn't they use "project" as the prefix? if they use "project." then our indices would end up as "project..$project.$uuid.$date"

I think we would want to discourage having a user provide the delimiter value in the case they don't provide it it could cause some grief

richm: Ok.  It makes the logic easier if the delimiter is included with the prefix - I can do things like
```
		return project_prefix + project + ".*";
```
as opposed to
```
    if (StringUtils.isEmtpy(project_prefix))
        return project + ".*";
    else
        return project_prefix + "." + project + ".*";
```

richm:
@ewolinetz @jcantrill fixed review comments - PTAL

ewolinetz:
LGTM

richm:
[merge]

richm:
[merge]

ewolinetz:
[merge]

PR #50, opened 2016-10-18
Allowing users to be able to view operations logs as cluster-reader i…

### LINKED ISSUE #42: Allow Cluster Reader to have access to the .all alias and .operations index
We should allow users the ability to configure their plugin to determine if just `cluster-admin` or `cluster-reader` as well, can see the infrastructure logs.

ewolinetz:
…f configured

Resolves #42 

@jcantrill @richm PTAL

### DIFF
--- a/src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
+++ b/src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
@@ -21,7 +21,6 @@ import java.io.FileReader;
 import java.io.IOException;
 import java.io.Reader;
 import java.net.InetSocketAddress;
-import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -68,10 +67,8 @@ import com.floragunn.searchguard.ssl.util.SSLConfigConstants;
 
 import io.fabric8.elasticsearch.plugin.ConfigurationSettings;
 import io.fabric8.kubernetes.client.ConfigBuilder;
+import io.fabric8.openshift.api.model.ClusterRoleBinding;
 import io.fabric8.openshift.api.model.Project;
-import io.fabric8.openshift.api.model.SubjectAccessReview;
-import io.fabric8.openshift.api.model.SubjectAccessReviewBuilder;
-import io.fabric8.openshift.api.model.SubjectAccessReviewResponse;
 import io.fabric8.openshift.client.DefaultOpenShiftClient;
 import io.fabric8.openshift.client.OpenShiftClient;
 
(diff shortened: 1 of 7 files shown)


richm:
lgtm but I don't know the openshift api

jcantrill:
LGTM assuming 'tokenreview' is enough to indicate a user is a clusterreader.

jcantrill:
After reading @liggit response, apparently we should revisit how we determine if a user is a cluster-admin since we are trying to 'educate dolphins'  ...which would probably be pretty easy since they are smart animals

ewolinetz:
@jcantrill updated pattern for checking roles, we will need to update our deployer to let the ES serviceaccount be able to 'get' 'clusterrolebindings'.  `oadm policy add-cluster-role-to-user {the-new-role} {es-serviceaccount-user}`

PTAL

src/main/java/io/fabric8/elasticsearch/plugin/acl/SearchGuardRolesMapping.java
```diff
@@ -96,7 +96,7 @@ public void syncFrom(UserProjectCache cache, final String userProfilePrefix){
 				builder.addUser(projectRoleName, username);
 			}
 			
-			if ( cache.isClusterAdmin(username) ) {
+			if ( cache.isOperationsUser(username) ) {
```
jcantrill: For our needs, is an operations user who could be a cluster-admin or cluster-reader the same 'ADMIN_ROLE'?  Do we have concerns that a cluster-reader may have access to logs they should not?

src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
```diff
@@ -270,25 +267,35 @@ private boolean isBlacklistProject(String project) {
 		return ArrayUtils.contains(operationsProjects, project.toLowerCase());
 	}
 
-	private boolean isClusterAdmin(final String token){
-		ConfigBuilder builder = new ConfigBuilder()
-				.withOauthToken(token);
+	private boolean isOperationsUser(final String username){
+		ConfigBuilder builder = new ConfigBuilder();
+		OpenShiftClient osClient = new DefaultOpenShiftClient(builder.build());
```
jcantrill: I think this class is 'Closeable' which means you can:
```
try(OpenShiftClient osClient = new DefaultOpenShiftClient(builder.build())){
}catch(Exception){}
```
and not even need to explicit close the client

src/main/java/io/fabric8/elasticsearch/plugin/acl/UserProjectCache.java
```diff
@@ -29,7 +29,7 @@
 	 * @param project  the project
 	 * @param binding  the binding to get user info
 	 */
-	void update(final String user, Set<String> projects, boolean clusterAdmin);
+	void update(final String user, Set<String> projects, boolean operationsUser);
```
jcantrill: Please update the param comments

src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
ewolinetz: it flags it if we explicitly don't close it... i'll wrap it though

ewolinetz:
@jcantrill addressed comments

jcantrill:
LGTM

ewolinetz:
[merge]

ewolinetz:
@jcantrill can we do a release with this please?

jcantrill:
released as part of 2.3.5.7

Issue #49, opened 2016-10-17
Read index-pattern data from files rather than hardcoded constants


jcantrill:
as mentioned in #48

richm:
The kibana index-pattern data is a single line, several thousand byte JSON blob hardcoded into constant strings in the DocumentBuilder class.  Instead, we should read this data from JSON or YAML files.  viaq can generate the elasticsearch index templates as json or yaml files - it could also generate the corresponding kibana index-pattern data and other config data.  https://github.com/ViaQ/elasticsearch-templates/issues/17

richm:
Partially addressed by https://github.com/fabric8io/openshift-elasticsearch-plugin/pull/51 - if using elasticsearch index templates, kibana index patterns are not needed

jcantrill:
Closing as will be fixed by #65

PR #51, opened 2016-10-18
common data model does not require index patterns

richm:
When using the common data model with Elasticsearch index templates,
Kibana can use the index template data for the field index patterns,
so we do not need to load them here.


richm:
@jcantrill  @ewolinetz PTAL - This is related to https://github.com/fabric8io/openshift-elasticsearch-plugin/issues/49 - if we use index templates, we do not need to load Kibana index patterns from this plugin, so the need to fix that issue is much less urgent.

jcantrill:
LGTM

src/main/java/io/fabric8/elasticsearch/plugin/kibana/DocumentBuilder.java
```diff
@@ -60,20 +56,20 @@ public DocumentBuilder intervalName(String name) {
 	
 	public DocumentBuilder applicationFields(boolean use_cdm) {
 		if ( use_cdm )
-			return addField(FIELDS_FIELD, StringEscapeUtils.escapeJava(APPLICATION_CDM_FIELDS));
```
ewolinetz: these are the fields for the Kibana index mapping. this can be omitted and it'll still work?

richm: Yes, surprisingly.  Kibana will use the index mappings instead.

richm: Actually, these are what are called "Kibana index patterns".  If they are omitted, Kibana will see if the index/index pattern has index mappings defined and use them instead.

ewolinetz:
[merge]

richm:
@jcantrill @ewolinetz can we do a release, or are we waiting for https://github.com/fabric8io/openshift-elasticsearch-plugin/pull/50 too?

jcantrill:
released 2.3.5.6: https://fabric8-ci.fusesource.com/view/release%20builds/job/openshift-elasticsearch-plugin-release/29/

PR #55, opened 2016-12-19
[BZ-1371200] Warn on NoShardAvailableActionException instead of dumpi…

jcantrill:
…ng stack

This PR adds logs a warning when ACL is not seeded and shards are not available.

cc @ewolinetz @jimmidyson @richm 

[test]


ewolinetz:
I'm not sure if there are any scenarios where we wouldn't want to mute this exception... otherwise LGTM

jcantrill:
[merge]

PR #59, opened 2017-02-03
fixes 1416210. Check for existence of index before trying to create

jcantrill:
I am unable to reproduce locally and I'm guessing this is the only place where it could happen.  Searching through the ES code shows that  it will throw such an exception only on creates.


ewolinetz:
code LGTM

PR #65, opened 2017-02-24
fix #63. Make kibana index patterns loadable from external file

### LINKED ISSUE #63: Make Kibana Index Patterns externally configurable
Modify plugin to allow the patterns to be externally loadable.

https://github.com/fabric8io/openshift-elasticsearch-plugin/blob/master/src/main/java/io/fabric8/elasticsearch/plugin/kibana/DocumentBuilder.java#L22

* Move constant to external file
* Figure out how a user could provide a pointer to a file to load instead.

jcantrill:
This provides the ability to seed the kibana index patterns from files or default.

cc @ewolinetz @richm


pom.xml
```diff
@@ -148,7 +148,6 @@
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
       <version>2.1</version>
-      <scope>test</scope>
```
jcantrill: @richm I have concerts here that this library may not be packaged already

richm: @jcantrill packaged upstream or downstream?

src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
```diff
@@ -107,12 +107,14 @@
 	private Boolean enabled;
 	private Boolean seeded;
 
-	private final boolean use_cdm;
 	private final String cdm_project_prefix;
 
+    private KibanaSeed kibanaSeed;
```
richm: indentation

src/main/java/io/fabric8/elasticsearch/plugin/kibana/IndexMappingLoader.java
```diff
@@ -0,0 +1,84 @@
...
+    private final String opsMappingsTemplate;
+    private final String emptyProjectMappingsTemplate;
+    
+    @Inject
+    public IndexMappingLoader(final Settings settings) {
+        try {
+            
+        }catch(Exception e){
+            
+        }
```
richm: ?

src/main/java/io/fabric8/elasticsearch/plugin/kibana/IndexMappingLoader.java
```diff
@@ -0,0 +1,84 @@
...
+    
+    private String loadMapping(final Settings settings, final String key, final String keyDefault) {
+        String mapping = settings.get(key);
+        if(mapping != null && new File(mapping).exists()) {
+            logger.info("Trying to load Kibana mapping for {} from plugin: {}", key, mapping);
+            try {
+                InputStream stream = new FileInputStream(mapping.toString());
+                return IOUtils.toString(stream);
+            }catch(Exception e) {
+                logger.error("Unable to load the Kibana mapping specified by {}: {}", key, mapping);
```
richm: Does this also log the information from the Exception?

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
```diff
@@ -69,13 +70,20 @@
 	private static final String[] OPERATIONS_ROLES = {"operations-user"};
 	
 	public static final String DEFAULT_INDEX_FIELD = "defaultIndex";
+
+    private final IndexMappingLoader mappingLoader;
```
richm: indentation?

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
```diff
@@ -350,22 +358,20 @@ private static void remove(String user, Set<String> projects, Client esClient, S
...
+		String source;
 		if ( project.equalsIgnoreCase(OPERATIONS_PROJECT) || project.equalsIgnoreCase(ADMIN_ALIAS_NAME) )
-			sourceBuilder.operationsFields(use_cdm);
+			source = mappingLoader.getOperationsMappingsTemplate();
 		else if ( project.equalsIgnoreCase(BLANK_PROJECT) )
-			sourceBuilder.blankFields();
+			source = mappingLoader.getEmptyProjectMappingsTemplate();
 		else
-			sourceBuilder.applicationFields(use_cdm);
+		    source = mappingLoader.getApplicationMappingsTemplate();
```
richm: indentation

src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
richm: Do you need this to be a member of the class?  You only reference it in one place, in the same context where it is used.

src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
```diff
@@ -107,12 +107,14 @@
...
+
 	@Inject
-	public DynamicACLFilter(final UserProjectCache cache, final Settings settings, final Client client){
+	public DynamicACLFilter(final UserProjectCache cache, final Settings settings, final Client client, final KibanaSeed seed){
 		this.cache = cache;
+		this.kibanaSeed = seed;
```
richm: I don't think you need this.

src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
```diff
@@ -258,7 +259,7 @@ private boolean updateCache(final String user, final String token, final boolean
 			if (isOperationsUser)
 				roles.add("operations-user");
 
-			setDashboards(user, projects, roles, esClient, kibanaIndex, kbnVersion, use_cdm, cdm_project_prefix, settings);
+			kibanaSeed.setDashboards(user, projects, roles, esClient, kibanaIndex, kbnVersion, cdm_project_prefix, settings);
```
richm: Why not just use `seed`, the function parameter?

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
```diff
@@ -350,22 +358,20 @@ private static void remove(String user, Set<String> projects, Client esClient, S
...
 		
-		String source = sourceBuilder.build();
+		source = source.replaceAll("$TITLE$", indexPattern);
```
richm: The `mappingLoader.get*MappingsTemplate` functions can return `null` - how will you prevent NPE?

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
```diff
@@ -69,13 +70,20 @@
 	private static final String[] OPERATIONS_ROLES = {"operations-user"};
 	
 	public static final String DEFAULT_INDEX_FIELD = "defaultIndex";
+
+    private final IndexMappingLoader mappingLoader;
+	
+	@Inject
+	public KibanaSeed(IndexMappingLoader loader) {
```
richm: where is this constructor used?

src/main/java/io/fabric8/elasticsearch/plugin/kibana/IndexMappingLoader.java
```diff
@@ -0,0 +1,84 @@
...
+    public static final String DEFAULT_OPERATIONS_MAPPING = "io/fabric8/elasticsearch/plugin/kibana/operations_mapping.json";
+    public static final String DEFAULT_APPLICATIONS_MAPPING = "io/fabric8/elasticsearch/plugin/kibana/applications_mapping.json";
+    public static final String DEFAULT_EMPTY_MAPPING = "io/fabric8/elasticsearch/plugin/kibana/empty_project_mappings.json";
+    private static ESLogger logger = Loggers.getLogger(IndexMappingLoader.class);
+    private final String appMappingsTemplate;
+    private final String opsMappingsTemplate;
+    private final String emptyProjectMappingsTemplate;
+    
+    @Inject
+    public IndexMappingLoader(final Settings settings) {
```
richm: where is this constructor called?

richm:
How will the deployer (ansible?) or the elasticsearch pod install.sh?  run.sh?  provide the index mapping files?

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
ewolinetz: `@Inject` is magic; this constructor will be called when we are expecting it in the DynamicACL constructor

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
```diff
@@ -350,22 +358,20 @@ private static void remove(String user, Set<String> projects, Client esClient, S
...
-			sourceBuilder.operationsFields(use_cdm);
+			source = mappingLoader.getOperationsMappingsTemplate();
 		else if ( project.equalsIgnoreCase(BLANK_PROJECT) )
-			sourceBuilder.blankFields();
+			source = mappingLoader.getEmptyProjectMappingsTemplate();
 		else
-			sourceBuilder.applicationFields(use_cdm);
+		    source = mappingLoader.getApplicationMappingsTemplate();
 		
-		String source = sourceBuilder.build();
```
ewolinetz: why wouldn't we have the mappingLoader be called from the doc builder?

src/main/resources/io/fabric8/elasticsearch/plugin/kibana/applications_mapping.json
```diff
@@ -0,0 +1,6 @@
+{
+   "title":"$TITLE$",
+   "timeFieldName":"time",
```
ewolinetz: shouldn't we use the time_field_name that's provided to createIndex?

src/main/resources/io/fabric8/elasticsearch/plugin/kibana/empty_project_mappings.json
```diff
@@ -0,0 +1,6 @@
+{
+   "title":"$TITLE$",
+   "timeFieldName":"time",
+   "description":"This is the Kibana index file for empty projects",
+   "fields": "[{\"name\":\"_index\",\"type\":\"string\",\"count\":0,\"scripted\":false,\"indexed\":false,\"analyzed\":false,\"doc_values\":false},        {\"name\":\"_type\",\"type\":\"string\",\"count\":0,\"scripted\":false,\"indexed\":true,\"analyzed\":false,\"doc_values\":false},{\"name\":\"_source\",\"type\":\"_source\",\"count\":0,\"scripted\":false,\"indexed\":false,\"analyzed\":false,\"doc_values\":false},{\"name\":\"_id\",\"type\":\"string\",\"count\":0,\"scripted\":false,\"indexed\":false,\"analyzed\":false,\"doc_values\":false},{\"name\":\"time\",\"type\":\"date\",\"count\":0,\"scripted\":false,\"indexed\":true,\"analyzed\":false,\"doc_values\":false}\r\n    ]"
```
ewolinetz: can we use single quotes to wrap the value and then we don't need to escape the double quotes?

src/main/java/io/fabric8/elasticsearch/plugin/kibana/IndexMappingLoader.java
jcantrill: removed.

src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
jcantrill: I converted from a Static utility class to an instance and it is used to seed dashboards.  Reformatted the class.

src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
jcantrill: Not sure of what you are refering to here

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
jcantrill: We should never have a null since we default from the classpath, but I added a null check to guard

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
jcantrill: I removed it since the doc builder builds up a document based on various calls to produces the desired outcome.  This change requires the document to be provided complete so there is not a document 'to build'.

src/main/resources/io/fabric8/elasticsearch/plugin/kibana/applications_mapping.json
jcantrill: If we are providing the complete document there is no need to use what is provided in the message arg.

src/main/resources/io/fabric8/elasticsearch/plugin/kibana/empty_project_mappings.json
jcantrill: Tried single quotes and Kibana doesn't seem to want to parse it.  I'm leaving for now.

jcantrill:
@richm @ewolinetz  Updated based on feedback.... I apologize for the formatting bits that made the changes less obvious.

pom.xml
jcantrill: downstream for enterprise.

src/main/resources/io/fabric8/elasticsearch/plugin/kibana/applications_mapping.json
ewolinetz: right, but when we call the 'create' method in KibanaSeed, we pass in a 'time_field_name' string variable. Why would we not use the value passed in instead of hardcoding "time" here?

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
ewolinetz: Understood, but we end up requiring extra logic/processing to then update the value within the file so that we can insert the index name. We are missing that for the time field name

src/main/resources/io/fabric8/elasticsearch/plugin/kibana/applications_mapping.json
jcantrill: This is the value directly exported out of Kibana for this index.  The mapping document should be provided 'mostly' complete.  It's a field for applications and I don't  see the point in substituting it in when it can just be provided as needed by whom ever is providing the document. I will update the class to remove 'time_field_name'

jcantrill:
@richm @ewolinetz updated to remove time field parameter.  Updated Readme

ewolinetz:
Why was the time field parameter removed? it is different whether we use the CDM model or not, we shouldn't be hardcoding it to be 'time' since it can also be '@timestamp'

richm:
@ewolinetz the time field name now comes from the index pattern json file.  The default is "time".  If the user specifies a common data model index pattern, it will use "@timestamp".

jcantrill:
@ewolinetz what @richm said, the value is now in the file instead of progammatically added by building up the document.  With that said, does the default value in the file need to be modified to '@timestamp'.  The defaults I added were from me pulling them straight from the source kibana index which said 'time'.  If that is not correct then we should alter it.

richm:
@jcantrill the default should be 'time'

jcantrill:
[merge] based on irc approval from @sosiouxme

jcantrill:
[merge]

jcantrill:
[merge]

jcantrill:
[merge]

PR #66, opened 2017-02-27
bz 1427254  Kibana Index Mode

jcantrill:
This is a result of a discussion I had with @portante and ops regarding https://bugzilla.redhat.com/show_bug.cgi?id=1427254

Justification is ops users already have access to the entire cluster and this would allow them to share dashboards and visualizations unrestricted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,13 +2,28 @@
 This is an OpenShift plugin to ElasticSearch to:
 
 * Dynamically update the SearchGuard ACL based on a user's name
-* Transform kibana index requests to support multitenant deployments
+* Transform kibana index requests to support multitenant deployments for 
+  non-operations users when so configured.
+
+*Note:*
+Previous versions of this plugin created a Kibana profile for each user regardless of their role,
+which is still the default mode of operation.  It is now possible to configure the Kibana index mode 
+to allow [operations users](./src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java) to share the Kibana index to store dashboards and visualizations.  It is highly 
+recommended that operations teams establish agreements and naming conventions so users do not 
+overwrite each others work.
 
 ## Configuring your initial ACLs
 With the update to use Searchguard-2 and Searchguard-SSL for ES 2.4.x, the
 OpenShift-Elasticsearch-Plugin assumes the initial ACLs are seeded when the cluster is started.
 
-You can view sample configurations [here] (./samples/).
+### Action Groups
+
+This plugin will produce role and role mappings that assume certain action groups are defined with
+the base SearchGuard ACL documents.  [Samples](./src/test/resources/io/fabric8/elasticsearch/plugin/roles_ops_shared_kibana_index_with_unique.yml) 
+of what the generated roles are found with the unit tests
+
+
+You can view sample ACL document configurations [here](./samples/).
 
 As with `sgadmin`, the plugin needs to use the certificate with a DN that matches
 the `searchguard.authcz.admin_dn` as defined in the ES config to be able to
@@ -44,11 +59,9 @@ The following additional parameters can be set in set in `elasticsearch.yml`:
 |*io.fabric8.elasticsearch.acl.user_profile_prefix*| The prefix to use to store Kibana user visualizations (default: `.kibana.USERUUID`)|
 |*io.fabric8.elasticsearch.kibana.mapping.app*| Absolute file path to a JSON document that defines the index mapping for applications| 
 |*io.fabric8.elasticsearch.kibana.mapping.ops*| Absolute file path to a JSON document that defines the index mapping for operations|
-|*io.fabric8.elasticsearch.kibana.mapping.empty| Absolute file path to a JSON document that defines the index mapping for blank indexes|
-|*openshift.config.project_index_prefix*| The string value that project/namespace indices use as their prefix (default: ``) for example, with the
-  common data model, if the namespace is `test`, the index name will be
-  `project.test.$uuid.YYYY.MM.DD`.  In this case, use `"project"` as the
-  prefix - do not include the trailing `.`.|
+|*io.fabric8.elasticsearch.kibana.mapping.empty*| Absolute file path to a JSON document that defines the index mapping for blank indexes|
+|*openshift.config.project_index_prefix*| The string value that project/namespace indices use as their prefix (default: ``) for example, with the common data model, if the namespace is `test`, the index name will be `project.test.$uuid.YYYY.MM.DD`.  In this case, use `"project"` as the  prefix - do not include the trailing `.`.|
+|*openshift.kibana.index.mode*| The setting that determines the kibana index is used by users.  Valid values are one of the following: <ul><li>**unique** (Default) - Each user gets a unique index for kibana visualizations (e.g. .kibana.USER_UUID)</ul></li><ul><li>**ops_shared**       - Users who are in an ops role will share an index (e.g. kibana) while non ops users will have a unique index (e.g. .kibana.USER_UUID)</ul></li>|
 
 *Note*: The `io.fabric8.elasticsearch.kibana.mapping.*` properties are required and must be defined for the plugin to function. A sample file
 may be found in the `samples` folder.
--- a/pom.xml
+++ b/pom.xml
@@ -86,16 +86,6 @@
       <artifactId>elasticsearch</artifactId>
       <version>${elasticsearch.version}</version>
       <scope>provided</scope>
-      <exclusions>
-      	<exclusion>
-      		<artifactId>jackson-core</artifactId>
-      		<groupId>com.fasterxml.jackson.core</groupId>
-      	</exclusion>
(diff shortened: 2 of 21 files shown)


richm:
This means all ops users are the same user.  This means if an admin makes customizations, someone else can overwrite those customizations?  All ops users are _not_ the same actual user.  There may be customers who will be perturbed because some other admin can wipe out their customizations.

portante:
@richm, ops users can still create their own dashboards with a different name.  We have hundreds of dashboards on our Kibana instance.  Is it not reasonable to assume that what one ops user needs can be shared by all, and they can agree using dashboard names to avoid conflicts?

richm:
@portante ok - but that means admins will need some agreement and convention about which dashboards are shared, and which are private, rather than being strictly separated by username/userid.  And that will not be consistent with the way normal users have their own config/dashboards.

jcantrill:
@richm Agree all ops users are not the same but most likely all ops users will require visualizations to the same data.  As @portante  suggested, they can come to an agreement on a naming convention.  Aside from figuring out how to 'share' across our design, this is the best scenario to enable sharable dashboards for ops.  The only thing we might add is a config setting that allows you to make it work in a legacy fashion, but generally I see this as a reasonable change.

richm:
Proposed documentation:
```
There is a change from previous versions.
If you login to Kibana as an administrative user, **all of your configs and dashboards are shared with all other administrative users**.  You will need to come up with some sort of naming convention and agreement to share and not overwrite the work of other admin users.
This does not apply if you login to Kibana as a non-admin user.  All of your configs and dashboards will continue to be private to your userid.
```

jcantrill:
One drawback I can think of for this arrangement is users who might want to separate their 'admin' experience from their 'user' experience; they will ONLY be able to use their 'user' profile if some one removes their cluster-admin role.  Do we care? Is this a concern?

portante:
@richm, perhaps explain that this facilitates sharing of dashboards across admin users who are able to see all the data anyways, while preserving the nature of multi-tenancy for non-admin users.  "With great power comes great responsibility."

richm:
There are more elegant ways to share configs/dashboards.  However, if this functionality needed urgently, we can go with this fix and revisit this issue at a later date.

jcantrill:
@richm @portante  I fixed up but will need to rebase once #66 merges. Everything is all jumbled here.

jimmidyson:
retest this please

jcantrill:
[test]

jimmidyson:
retest this please

jimmidyson:
@jcantrill Tests passed now that AWS has sorted stuff out. Will leave it to you to merge.

README.md
```diff
@@ -2,9 +2,16 @@
 This is an OpenShift plugin to ElasticSearch to:
 
 * Dynamically update the SearchGuard ACL based on a user's name
-* Transform kibana index requests to support multitenant deployments
+* Transform kibana index requests to support multitenant deployments for non-operations users.
 * Seed the Searchguard index `config`, `roles`, `rolesmapping`, and `actiongroups` types
 
+*Note:*
+Previous versions of this plugin created a Kibana profile for each user regardless of their role.
+As of v2.4.4.2, any users that has been given the OpenShift cluster-admin role will share the same
```
portante: s/has/have

pom.xml
```diff
@@ -79,7 +79,7 @@
     <!-- used by vendor builds to use vendor version of searchguard plugins -->
     <version.suffix />
   </properties>
-
+
```
portante: Seems like an unnecessary change ...

src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
```diff
@@ -78,407 +78,424 @@
...
-		String kspass = settings.get(SG_CLIENT_KS_PASS, DEFAULT_SG_CLIENT_KS_PASS);
-		String tspass = settings.get(SG_CLIENT_TS_PASS, DEFAULT_SG_CLIENT_TS_PASS);
-		String kstype = settings.get(SG_CLIENT_KS_TYPE, DEFAULT_SG_CLIENT_KS_TYPE);
-		String tstype = settings.get(SG_CLIENT_TS_TYPE, DEFAULT_SG_CLIENT_TS_TYPE);
-
-		Settings.Builder settingsBuilder = Settings
-                .builder()
-                .put("path.home", ".")
-                .put("path.conf", ".")
+public class DynamicACLFilter extends RestFilter implements ConfigurationSettings {
```
portante: This looks like a huge change just for whitespace mostly.  Why such a large amount of change like that?

README.md
```diff
@@ -2,9 +2,16 @@
...
+Kibana index to store configurations and dashboards.  This change makes it possible to share dashboards
+between operation users.  It is highly recommended that operations teams establish agreements and
+naming conventions so users do not overwrite each others work.
```
portante: So we need to consider this change a bit in terms of how one would migrate to start using it.

What happens if we deploy this where each operation user has built up a set of dashboards already?  What happens to access to those previously created dashboards?

It would be great to have an "opt-in" switch to turn it on, so that admins can prepare by first exporting the dashboards they want.

Even with the opt-in switch it would be great to be able to provide a tool to dump each ops user's individual kibana dashboard so that they could be imported to the common one.

portante: We also should dove tail this work with the unified dashboard work for non-ops users.

jcantrill:
@richm @portante  This is a reworked PR for https://bugzilla.redhat.com/show_bug.cgi?id=1427254.

I pushed a test version of the image to: docker.io/jcantrill/logging-elasticsearch:bz1427254

which can be utilized by configuring the logging-elasticsearch configmap with:
`openshift.kibana.index.mode: ops_shared`

There are a number of refactorings to make this work.  I think we will additionally need the following:

*operations and .all alias dont seem to map properly. I propose to fix that in subsequent PR
* I think we need to modify hwo the roles are generated per what @richm  found.  User roles to all their projects.  I think this PR will be less then useful. I propose fixing that in subsequent PR.

Please test to see if we think it works as we desire.

jcantrill:
@portante Updated to only 2 modes per your suggestion.  @richm the other change this brings in is to move the allowed actions to action groups defined in the config files.  This will allow us to tweak actions without having to rebuild the plugin for the given roles.  We will need a logging PR to set the defaults and pull in this change.

README.md
```diff
@@ -2,13 +2,28 @@
...
 
 * Dynamically update the SearchGuard ACL based on a user's name
-* Transform kibana index requests to support multitenant deployments
+* Transform kibana index requests to support multitenant deployments for 
+  non-operations users when so configured.
+
+*Note:*
+Previous versions of this plugin created a Kibana profile for each user regardless of their role,
+which is still the default mode of operation.  It is now possible to configure the Kibana index mode 
+to allow operations users to share the Kibana index to store dashboards and visualizations.  It is highly
```
portante: Is it worth defining what defines an "operations" user is?

src/main/java/io/fabric8/elasticsearch/plugin/KibanaIndexMode.java
```diff
@@ -0,0 +1,31 @@
...
+     * The setting that determines the kibana index is used by users.  Valid values are one of the following:
+     * 
+     *   * unique (Default) - Each user gets a unique index for kibana visualizations (e.g. .kibana.USER_UUID)
+     *   * ops_shared       - Users who are in an ops role will share an index (e.g. kibana) while non ops users will 
+     *                        have a unique index (e.g. .kibana.USER_UUID)
+     */
+    static final String OPENSHIFT_KIBANA_INDEX_MODE = "openshift.kibana.index.mode";
+    
+    static final String UNIQUE = "unique";
+    static final String SHARED_OPS = "ops_shared";
```
portante: Do you care the constant is `SHARED_OPS` instead of `OPS_SHARED`?

README.md
jcantrill: Added link to the file where the SAR is used.

src/main/java/io/fabric8/elasticsearch/plugin/KibanaIndexMode.java
jcantrill: modified value to 'shared_ops'

richm:
lgtm, but I don't quite follow all of the java changes

jcantrill:
[(3) Ability to seed Kibana dashboard for admins[logging][epic-ois-agl-exp]](https://trello.com/c/O6MCrGUx/454-3-ability-to-seed-kibana-dashboard-for-adminsloggingepic-ois-agl-exp)

jcantrill:
[(3) Ability to seed Kibana dashboard for admins[logging][epic-ois-agl-exp]](https://trello.com/c/O6MCrGUx/454-3-ability-to-seed-kibana-dashboard-for-adminsloggingepic-ois-agl-exp)

PR #73, opened 2017-05-04
bug 1446217. Use SAR to determine isOperationalUser

### LINKED ISSUE #72: Invalid method for determining the user's permissions
The [isOperationsUser](https://github.com/fabric8io/openshift-elasticsearch-plugin/blob/7291d132c3464610921f93d3a8c96453001d6bfd/src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java#L278-L296) method incorrectly assumes that a user must be directly bound to `cluster-admins` or `cluster-readers` to see the logs.  This is incorrect in two scenarios:

1. The user has rights due to a different binding
2. The user belongs to a group that grants the rights

Instead, the method should perform a [SubjectAccessReview](https://docs.openshift.com/container-platform/latest/rest_api/openshift_v1.html#v1-subjectaccessreview) based on the username and group information (you will probably need to build a reverse index of groups).  The following `SAR` is equivalent to cluster admin:

```json
{
	"kind" : "SubjectAccessReview",
	"apiVersion" : "v1",
	"user" : "<username>",
	"groups" : ["<groups>"],
	"verb" : "*",
	"resource" : "*",
}
```

The following `SAR` could serve as a proxy for `cluster-reader` but that is up to you guys:

```json
{
	"kind" : "SubjectAccessReview",
	"apiVersion" : "v1",
	"user" : "<username>",
	"groups" : ["<groups>"],
	"verb" : "list",
	"resource" : "pods/log",
}
```

cc  @ewolinetz @jcantrill @richm 

@gabemontero may be able to provide guidance on how to perform a `SAR` in Java.

jcantrill:
fixes #72

### DIFF
--- a/src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
+++ b/src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
@@ -57,9 +57,10 @@ import io.fabric8.elasticsearch.plugin.ConfigurationSettings;
 import io.fabric8.elasticsearch.plugin.kibana.KibanaSeed;
 import io.fabric8.kubernetes.client.ConfigBuilder;
 import io.fabric8.kubernetes.client.KubernetesClientException;
-import io.fabric8.openshift.api.model.ClusterRoleBinding;
 import io.fabric8.openshift.api.model.Project;
+import io.fabric8.openshift.api.model.SubjectAccessReviewResponse;
 import io.fabric8.openshift.client.DefaultOpenShiftClient;
+import io.fabric8.openshift.client.NamespacedOpenShiftClient;
 import io.fabric8.openshift.client.OpenShiftClient;
 
 /**
@@ -176,7 +177,7 @@ public class DynamicACLFilter extends RestFilter implements ConfigurationSetting
                     logger.debug("Cache has user: {}", cache.hasUser(user, token));
                 }
                 if (StringUtils.isNotEmpty(token) && StringUtils.isNotEmpty(user) && !cache.hasUser(user, token)) {
-                    final boolean isOperationsUser = isOperationsUser(user);
+                    final boolean isOperationsUser = isOperationsUser(user, token);
                     if (isOperationsUser) {
                         request.putInContext(OPENSHIFT_ROLES, "cluster-admin");
                     }
@@ -269,36 +270,24 @@ public class DynamicACLFilter extends RestFilter implements ConfigurationSetting
         return ArrayUtils.contains(operationsProjects, project.toLowerCase());
     }
 
-    private boolean isOperationsUser(final String username) {
-        ConfigBuilder builder = new ConfigBuilder();
-
-        boolean clusterReaderAllowed = settings.getAsBoolean(OPENSHIFT_ALLOW_CLUSTER_READER,
-                DEFAULT_OPENSHIFT_ALLOW_CLUSTER_READER);
-
-        try (OpenShiftClient osClient = new DefaultOpenShiftClient(builder.build())) {
-            ClusterRoleBinding bResponse = osClient.clusterRoleBindings().inNamespace("").withName("cluster-admins")
-                    .get();
-
-            logger.debug("Does '{}' exist in cluster-admins? '{}'", username,
-                    bResponse.getUserNames().contains(username));
-
-            if (clusterReaderAllowed) {
-                ClusterRoleBinding readerResponse = osClient.clusterRoleBindings().inNamespace("")
-                        .withName("cluster-readers").get();
-
-                logger.debug("Does '{}' exist in cluster-readers? '{}'", username,
-                        readerResponse.getUserNames().contains(username));
-
-                if (readerResponse.getUserNames().contains(username))
-                    return true;
-            }
-
-            return bResponse.getUserNames().contains(username);
+    private boolean isOperationsUser(final String user, final String token) {
+        ConfigBuilder builder = new ConfigBuilder().withOauthToken(token);
+        boolean allowed = false;
+        try (NamespacedOpenShiftClient osClient = new DefaultOpenShiftClient(builder.build())) {
+            logger.debug("Submitting a SAR to see if '{}' is able to retrieve logs across the cluster", user);
+            SubjectAccessReviewResponse response = osClient.inAnyNamespace()
+                    .subjectAccessReviews()
+                    .createNew()
+                    .withVerb("get")
+                    .withResource("pods/log")
+                    .done();
+            allowed = response.getAllowed();
         } catch (Exception e) {
-            logger.error("Exception determining user's role.", e);
+            logger.error("Exception determining user's '{}' role.", e, user);
+        }finally {
+            logger.debug("User '{}' isOperationsUser: {}", user, allowed);
         }
-
-        return false;
+        return allowed;
     }
 
     private void syncAcl() {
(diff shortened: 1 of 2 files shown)


jcantrill:
cc @portante

richm:
For what reason are you switching to use the token for the auth to perform the query?  Is it possible that the user is not allowed to perform the query, but is allowed to view the logs?

jcantrill:
I think if you have 'get' access across the operational namespaces that would mean you are allowed to view those namespaces.  And by query do you mean against ES, or via a SAR?  I otherwise defer to @deads2k and @liggitt

src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
```diff
@@ -269,36 +272,24 @@ private boolean isBlacklistProject(String project) {
...
-                    return true;
-            }
-
-            return bResponse.getUserNames().contains(username);
+    private boolean isOperationsUser(final String user, final String token) {
+        ConfigBuilder builder = new ConfigBuilder().withOauthToken(token);
+        boolean allowed = false;
+        try (NamespacedOpenShiftClient osClient = new DefaultOpenShiftClient(builder.build())) {
+            logger.debug("Submitting a SAR to see if '{}' is able to retrieve logs across the cluster", user);
+            SubjectAccessReviewResponse response = osClient.inAnyNamespace()
```
deads2k: In 3.6, you'll want to start using the newly available `/apis/authorization.k8s.io/v1/selfsubjectaccessreviews`.

deads2k:
> I think if you have 'get' access across the operational namespaces that would mean you are allowed to view those namespaces

If you have "get" on "pods/log" across all namespaces.

src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
```diff
@@ -269,36 +272,24 @@ private boolean isBlacklistProject(String project) {
...
+                    .subjectAccessReviews()
+                    .createNew()
+                    .withVerb("get")
+                    .withResource("pod/logs")
```
deads2k: `pods/log` I'd think. https://github.com/openshift/origin/blob/master/pkg/cmd/server/bootstrappolicy/policy.go#L158

richm:
@jcantrill yeah, I mean against openshift to read/validate/whatever the sar - can a user have permission to read/get/view pods/logs, but not have permission to read/validate/whatever the sar?

enj:
> can a user have permission to read/get/view pods/logs, but not have permission to read/validate/whatever the sar?

@richm The `self-access-reviewer` role is bound to everyone, and `basic-user` is bound to all authenticated users.  Both of these grant the self-SAR ability and both would have to altered and reconcile protected by the CA for this check to fail.  If that is a concern, you can change this to issue a SAR with the user's information, but you would have to determine what groups they belong to.

src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
richm: @deads2k or @jliggit - question about how sar works: what if there are namespaces A, B, and C - and only those namespaces - I have been manually granted read logs access to A, B, and C - will the sar come back and say 'yes, richm has access to read logs in all namespaces'?

richm: @enj same question as above ^^^

enj: No, the SAR will say you cannot access logs in all namespaces.  A cluster scoped binding with that rule has to exist, and in your example three local bindings exist.  In a way you can think of "all namespaces" as all namespaces that exist now and all that will exist in the future.

BTW Jordan's username is liggitt not jliggit (no j in the front and two t's at the end).

richm: Ok.  So the above code will answer the question - Does `token` have access to all logs in all namespaces?

enj: Correct, but @deads2k's comments about `pods/log` and `/apis/authorization.k8s.io/v1/selfsubjectaccessreviews` still stand.  The former should definitely be addressed, and the latter is important since it does not make sense for new code to use a legacy endpoint.

jcantrill:
merge

portante:
@jcantrill, seems like there are some questions still about `pods/log` vs `/pods/logs`, and a question about using the legacy end point.

jcantrill:
@portante I opened https://github.com/fabric8io/openshift-elasticsearch-plugin/issues/74 to address the legacy endpoint. I'm not sure exactly by looking at the referenced code how it is known that it is using legacy anyway.  Also, given the endpoint will exist for the foreseeable future and is supported on the version of Openshift we are targeting, I don't see it as a high priority.

Also, doesn't the responses provided by @enj address the open questions?

portante:
@jcantrill, the last comment from @enj is an hour ago (at this point in time), is that what you see?  Do you see any other comments since, or changes since, which address the two topics?

enj:
I think everything has been addressed expect #74.  One important thing to consider is that the legacy endpoint is likely to be slower than the new one (because we do a bunch of checks to map it internally into the correct/new endpoint).

jcantrill:
@iocanel  Can you tell me how to use the apigroups endpoint for SAR or confirm this one is using it?

enj:
> confirm this one is using it

@jcantrill If it was using the new endpoint it would likely be called `selfSubjectAccessReviews` instead of `subjectAccessReviews`.

ewolinetz:
I'm a little hesitant to completely rip out the method of letting access be determined by the `cluster-admin` and `cluster-reader` roles as that was a previous RFE to be able to do that. Perhaps this can be something you can opt out of if you'd like to continue using the current means of restricting access to the .operations index?

I'd like to hear from someone on the ops team too regarding this. @wshearn ?

enj:
@ewolinetz If anything, this should let more users see the logs, not less (both cluster admins and cluster readers will pass the self-SAR, and thus will continue to be able to see the logs).  The only thing you have lost is the ability to require cluster admin privileges to see said logs.  Also, there is no guarantee that the rules that are bound to those roles have not changed, so a self-SAR is a more accurate check.

liggitt:
> if you'd like to continue using the current means of restricting access to the .operations index

the current method is incorrect.

ewolinetz:
@enj right that's my concern and the reason I would like to hear from operations -- I don't want this to start giving users access when they normally would not. Specifically in the hosted environments.

I understand and agree that this is the more accurate way to do this, I would just like to verify this won't cause a security risk for them.

ewolinetz:
since we'd be removing the need to do rolebinding validation, we can also update the deployer documentation/template as we won't need to rolebinding-reader role anymore.

jcantrill:
@enj in light of we dont currently have a client to do a SelfSubjectReview, is it sufficient to address this once it is available.  Does this PR still provide us with a better security posture?

@ewolinetz have your concerns been addressed?

enj:
> in light of we dont currently have a client to do a SelfSubjectReview, is it sufficient to address this once it is available. Does this PR still provide us with a better security posture?

@jcantrill Yeah, I think it is fine to address it once you have it available.  This implementation is correct, even if it does not use the new endpoint.

richm:
Is this the same as `oc policy can-i $verb $resource`?

enj:
> Is this the same as `oc policy can-i $verb $resource`?

@richm The `can-i` command is a combination of `SelfSubjectRulesReviews` + `SubjectRulesReviews` + `SubjectAccessReviews`.  In your specific example, it does issue a self-SAR.

ewolinetz:
@jcantrill yes, i verified with wesley that there isn't an edge case that would cause a security concern

PR #76, opened 2017-05-11
fix #75. Replace time field and ensure JSON mapping files included in jar

### LINKED ISSUE #75: Use @timestamp instead of time field
The external kibana index pattern `.json` files make reference to the non-existent field "`time`" instead of the proper field "`@timestamp`".

It also appears that the jar file does not contain these `.json` files.

jcantrill:
@portante


richm:
What is this for?

src/main/resources/io/fabric8/elasticsearch/plugin/kibana/applications_mapping.json
```diff
@@ -1,6 +1,6 @@
...
...:true,\"doc_values\":false},    {\"name\":\"kubernetes_labels_deployment\",\"type\":\"string\",\"count\":0,\"scripted\":false,\"indexed\":true,\"analyzed\":true,\"doc_values\":false},    {\"name\":\"kubernetes_container_name\",\"type\":\"string\",\"count\":0,\"scripted\":false,\"indexed\":true,\"analyzed\":true,\"doc_values\":false},    {\"name\":\"kubernetes_pod_id\",\"type\":\"string\",\"count\":0,\"scripted\":false,\"indexed\":true,\"analyzed\":true,\"doc_values\":false},  {\"name\":\"kubernetes_labels_openshift_io/deployer-pod-for_name\",\"type\":\"string\",\"count\":0,\"scripted\":false,\"indexed\":true,\"analyzed\":true,\"doc_values\":false},    {\"name\":\"_source\",\"type\":\"_source\",\"count\":0,\"scripted\":false,\"indexed\":false,\"analyzed\":false,\"doc_values\":false},    {\"name\":\"_id\",\"type\":\"string\",\"count\":0,\"scripted\":false,\"indexed\":false,\"analyzed\":false,\"doc_values\":false},    {\"name\":\"@timestamp\",\"type\":\"date\",\"count\":0,\"scripted\":false,\"indexed\":true,\"analyzed\":false,\"doc_values\":false},{\"name\":\"kubernetes_host\",\"type\":\"string\",\"count\":0,\"scripted\":false,\"indexed\":true,\"analyzed\":true,\"doc_values\":false}]"
```
portante: whoa, these look like all the old namespace names!

portante:
@richm, we have a problem at a customer site where the latest code does not produce kibana index patterns that let you see logs.  Kibana gets an error in the javascript console looking for a "`time`" field that does not exist.  These defaults need to be updated to reference the new namespace for logging fields.

richm:
> Kibana gets an error in the javascript console looking for a "time" field that does not exist.

That's what https://github.com/fabric8io/openshift-elasticsearch-plugin/commit/71e4050701b6f6cb423bfa0737ab9db0ef8caf5a and https://github.com/openshift/origin-aggregated-logging/commit/eb83ac43a9c2dde14f747530facdf01133b35806 and https://github.com/openshift/openshift-ansible/commit/c4fcf9ef11b4dfc27f8f1e509719864f0ef2d019 are for

> These defaults need to be updated to reference the new namespace for logging fields.

Ok, but you lose the ability for this plugin to work with openshift 3.3 or earlier.  Perhaps that is ok.  I suppose you could support that by providing index pattern files for the old data model.

But this seems like a lot of work to solve a problem that is already solved.

portante:
@richm, so the problem we are having is that we have two environments, both of which have the external `index_patterns/*.json` files in the pod, both have no reference to those index pattern files in the `elasticsearch.yml` file, but one environment works, while the other doesn't.

Having the latest code default to the oldest behavior seems bad.  In fact, really, having any defaults in the plugin itself seems bad, as it cannot work on its own in all environments it supports.

So maybe we should take the defaults out entirely, throwing an error if the configuration files are not provided, refusing to load the plugin.  Then the users of the plugin in those environments need to provide the proper defaults.

How does that sound?

jcantrill:
@richm @portante removed the default index mappings which will require consumers to specify.  This will require a change to origin-aggregated-logging to define a setting for the 'empty-project' key.

portante:
This looks good to me.  So for any version of OpenShift which uses this version of the plugin, it needs to provide those mappings as arguments in the settings.

jcantrill:
Moved one of the mapping files to sample in case we forgot what it needs to look like.  Merging with @portante blessing.

PR #82, opened 2017-06-02
bug 1456413. Fix security error message for patterns with no indexes

jcantrill:
This PR fixes the error message displayed to users in kibana:

* Creates mappings only for indices that exist
* Modifies the response when fieldstat call is made

image


jcantrill:
cc @portante

PR #83, opened 2017-06-08
bug 1459221. Fix null pointer in KibanaUserReindexAction

jcantrill:
cc @portante

### DIFF
--- a/src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexAction.java
+++ b/src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexAction.java
@@ -36,8 +36,6 @@ import org.elasticsearch.action.support.ActionFilter;
 import org.elasticsearch.action.support.ActionFilterChain;
 import org.elasticsearch.client.Client;
 import org.elasticsearch.cluster.ClusterService;
-import org.elasticsearch.common.bytes.BytesArray;
-import org.elasticsearch.common.bytes.BytesReference;
 import org.elasticsearch.common.inject.Inject;
 import org.elasticsearch.common.io.stream.ByteBufferStreamInput;
 import org.elasticsearch.common.io.stream.BytesStreamOutput;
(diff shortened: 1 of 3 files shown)


src/main/java/io/fabric8/elasticsearch/plugin/KibanaUserReindexAction.java
```diff
@@ -168,38 +167,7 @@ public void apply(String action, ActionResponse response, @SuppressWarnings("raw
...
-        // Check for .kibana.* in the source
-        BytesReference replacedContent = null;
-        if (!response.isSourceEmpty()) {
-            String source = response.getSourceAsBytesRef().toUtf8();
-            String replaced = source.replaceAll(replacedIndex, index);
-            replacedContent = new BytesArray(replaced);
-        }
-
-        // Check for .kibana.* in the fields
-        Map<String, GetField> responseFields = response.getFields();
```
jcantrill: @portante The Nullpointer was here so it looks like response returns a null map.  I don't know under what condition this can occur. Its possible a bug in ES code, possibly a bug in SG code.  This refactoring essentially a defensive guard.

portante: @jcantrill, in the new code, the private variable `responseFields` is initialized to an empty `hashmap` at line 45 of `GetResultBuilder.java`, which avoid the `NullPointer` error at line 79 in `GetResultBuilder.java`, is that the correct understanding?

jcantrill: @portante yes

src/test/java/io/fabric8/elasticsearch/plugin/kibana/GetResultBuilderTest.java
```diff
@@ -0,0 +1,96 @@
...
+        this.response = mock(GetResponse.class);
+        
+        when(response.getType()).thenReturn("type");
+        when(response.getId()).thenReturn("id");
+        when(response.getVersion()).thenReturn(123L);
+        when(response.isExists()).thenReturn(true);
+    }
+    
+    private void givenAnInitializedBuilder() {
+        builder.response(response).index("foo").replacedIndex("bar");
```
richm: should this be `this.builder.response...`?  In Java, can you refer to private members without `this`?

jcantrill: java assumes you are referring to the current class and this is not required.

PR #85, opened 2017-06-14
project.* index pattern causes exception

richm:
If the user specifies an index pattern in Kibana like "project.*"
because they want to have a search for all projects that they
have access to, this will cause an exception in the openshift
elasticsearch plugin string handling.
The fix is to perform better string bounds checking.


richm:
@jcantrill PTAL

PR #94, opened 2017-07-23
bug 1473153. Always add .all for ops users

jcantrill:
bug 1473153. Always add .all for ops users

Only the latest commit is of interest.  The first commit should be merged before this PR.


jcantrill:
@lukas-vlcek

PR #95, opened 2017-07-28
modify roles to be user based so all of a users indexes are visible in a single query

jcantrill:
This PR introduces:
* ACL documents generated based on roles per user instead of per project [1]
* Resolves issue where user is unable to query across multiple indexes

The configuration option `openshift.acl.role_strategy` allows you to switch between generating roles per project where running queries across multiple project indices is not possible and generating roles per user.  The later creates proper role permissions to query across the indices you own.  I don't see us promoting or encouraging us of project role strategy going forward; this was a refactoring convenience to confirm everything was properly tested and backwards compatible.

[1] https://github.com/jcantrill/openshift-elasticsearch-plugin/blob/60f9d06520b73bcbd8bb74bfd17b71f89ac3d6ac/src/test/resources/io/fabric8/elasticsearch/plugin/user_role_with_shared_kibana_index_with_unique.yml

cc @pweil- @portante @richm

### DIFF
--- a/src/main/java/io/fabric8/elasticsearch/plugin/ConfigurationSettings.java
+++ b/src/main/java/io/fabric8/elasticsearch/plugin/ConfigurationSettings.java
@@ -87,6 +87,13 @@ public interface ConfigurationSettings extends KibanaIndexMode{
      * Kibana index.
      */
     static final String OPENSHIFT_DYNAMIC_ENABLED_FLAG = "openshift.acl.dynamic.enabled";
+    
+    /**
+     * The strategy to use for generating roles and role mappings
+     */
+    static final String OPENSHIFT_ACL_ROLE_STRATEGY = "openshift.acl.role_strategy";
+    static final String DEFAULT_ACL_ROLE_STRATEGY = "user";
+    
     static final String OPENSHIFT_KIBANA_REWRITE_ENABLED_FLAG = "openshift.kibana.rewrite.enabled";
 
     
--- a/src/main/java/io/fabric8/elasticsearch/plugin/PluginSettings.java
+++ b/src/main/java/io/fabric8/elasticsearch/plugin/PluginSettings.java
@@ -1,0 +1,102 @@
+/**
+ * Copyright (C) 2015 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.fabric8.elasticsearch.plugin;
+
+import static io.fabric8.elasticsearch.plugin.acl.SearchGuardSyncStrategyFactory.PROJECT;
+import static io.fabric8.elasticsearch.plugin.acl.SearchGuardSyncStrategyFactory.USER;
+
+import org.apache.commons.lang.ArrayUtils;
+import org.elasticsearch.common.inject.Inject;
+import org.elasticsearch.common.logging.ESLogger;
+import org.elasticsearch.common.logging.Loggers;
+import org.elasticsearch.common.settings.Settings;
+
+public class PluginSettings implements ConfigurationSettings {
+
+    private static final ESLogger LOGGER = Loggers.getLogger(PluginSettings.class);
+    
+    private String kibanaIndexMode;
+    private String roleStrategy;
+    private final String cdmProjectPrefix;
+    private final String defaultKibanaIndex;
+    private final String searchGuardIndex;
+    private final String kibanaVersion;
+    private final String kbnVersionHeader;
+    private final Boolean enabled;
+    
+    @Inject
+    public PluginSettings(final Settings settings) {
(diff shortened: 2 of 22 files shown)


jcantrill:
@ewolinetz  might you provide some feedback here?

src/main/java/io/fabric8/elasticsearch/plugin/acl/BaseRolesMappingSyncStrategy.java
```diff
@@ -0,0 +1,56 @@
...
+        this.mappings = mappings;
+    }
+
+
+    // Remove roles that start with "gen_"
+    private void removeSyncAcls() {
+        for (Iterator<RolesMapping> i = mappings.iterator(); i.hasNext();) {
+            RolesMapping mapping = i.next();
+            if (mapping.getName() != null && mapping.getName().startsWith(ROLE_PREFIX)) {
+                mappings.removeRolesMapping(mapping);
```
pweil-: wouldn't you call `remove` on the iterator here?

src/main/java/io/fabric8/elasticsearch/plugin/acl/BaseRolesSyncStrategy.java
```diff
@@ -0,0 +1,77 @@
...
+        syncFromImpl(cache, builder);
+        roles.addAll(builder.build());
+    }
+
+    // Remove roles that start with "gen_"
+    private void removeSyncAcls() {
+        for (Iterator<Roles> i = roles.iterator(); i.hasNext();) {
+            Roles role = i.next();
+            if (role.getName() != null && role.getName().startsWith(SearchGuardRoles.ROLE_PREFIX)) {
+                roles.removeRole(role);
```
pweil-: same question about usage of an interator but not using `remove`

ewolinetz:
@jcantrill could you update the description of this PR to further describe what changes you're proposing here? I'm not sure what `openshift.acl.role_strategy` is being used for and what its choices do.

src/main/java/io/fabric8/elasticsearch/plugin/acl/BaseRolesMappingSyncStrategy.java
jcantrill: We could, maybe, but:
1. The underlying impl doesnt ref the same collection.
1. This semantics allows us to modify the collection without concerns of it being modified elsewhere

src/main/java/io/fabric8/elasticsearch/plugin/acl/BaseRolesSyncStrategy.java
jcantrill: See https://github.com/fabric8io/openshift-elasticsearch-plugin/pull/95/files/60f9d06520b73bcbd8bb74bfd17b71f89ac3d6ac#r130917168

src/main/java/io/fabric8/elasticsearch/plugin/acl/BaseRolesMappingSyncStrategy.java
pweil-: If you're telling me it's safe then ok.  Usually you'd need to use `.remove()` because [the behavior of an iterator is unspecified if the underlying collection is modified while the iteration is in progress in any way other than by calling this method](https://docs.oracle.com/javase/8/docs/api/java/util/Iterator.html).  I don't have the code locally to dig in to so take that with a grain of salt

src/main/java/io/fabric8/elasticsearch/plugin/ConfigurationSettings.java
```diff
@@ -87,6 +87,12 @@
      * Kibana index.
      */
     static final String OPENSHIFT_DYNAMIC_ENABLED_FLAG = "openshift.acl.dynamic.enabled";
+    
+    /**
+     * The strategy to use for generating roles and role mappings
+     */
+    static final String OPENSHIFT_ACL_ROLE_STRATEGY = "openshift.acl.role_strategy";
```
ewolinetz: should we have a default value defined in the config?

src/main/java/io/fabric8/elasticsearch/plugin/PluginSettings.java
```diff
@@ -0,0 +1,102 @@
...
+    private final String cdmProjectPrefix;
+    private final String defaultKibanaIndex;
+    private final String searchGuardIndex;
+    private final String kibanaVersion;
+    private final String kbnVersionHeader;
+    private final Boolean enabled;
+    
+    @Inject
+    public PluginSettings(final Settings settings) {
+        this.kibanaIndexMode = settings.get(OPENSHIFT_KIBANA_INDEX_MODE, UNIQUE);
```
ewolinetz: use a config default instead here (matches already used pattern)

src/main/java/io/fabric8/elasticsearch/plugin/PluginSettings.java
```diff
@@ -0,0 +1,102 @@
...
+        if(!ArrayUtils.contains(new String [] {UNIQUE, SHARED_OPS}, kibanaIndexMode.toLowerCase())) {
+            this.kibanaIndexMode = UNIQUE;
+        }
+        
+        this.roleStrategy = settings.get(OPENSHIFT_ACL_ROLE_STRATEGY, USER);
```
ewolinetz: use a config default here?

src/main/java/io/fabric8/elasticsearch/plugin/ConfigurationSettings.java
jcantrill: I added a default to the config but I prefer not to advertise we have this option.

ewolinetz: +1 to that, i was thinking its more a convenience for us

PR #96, opened 2017-08-28
User dashboard share

FlorinPeter:
This PR will also allow users to share their dashboards.
The index mode "shared_ops" was also kept and a new mode "shared" was introduced.
In the shared mode users and ops will have different indices like ".kibana" and ".kibana_ops" for sharing dashboards.


src/main/java/io/fabric8/elasticsearch/plugin/KibanaIndexMode.java
```diff
@@ -21,12 +21,15 @@
...
+     *   * shared_ops       - Users who are in an ops role will share an index (e.g. kibana_ops) while non ops users will 
      *                        have a unique index (e.g. .kibana.USER_UUID)
+     *   * shared           - Users who are in an ops role will share an index (e.g. kibana_ops) while non ops users will 
+     *                        share the default index (e.g. .kibana)                      
      */
     static final String OPENSHIFT_KIBANA_INDEX_MODE = "openshift.kibana.index.mode";
     
     static final String UNIQUE = "unique";
     static final String SHARED_OPS = "shared_ops";
+    static final String SHARED = "shared";
```
jcantrill: We may wish to consider changing this name to something like 'shared_non_ops' in the event we wish to make kibana index 'shared' across all users.  That was my thought originally for 'shared'.  It would be interesting how a full 'shared' mode might exist given I assume we still might restrict non ops users to specific indicies

src/main/java/io/fabric8/elasticsearch/plugin/OpenshiftRequestContextFactory.java
```diff
@@ -63,7 +64,7 @@ public OpenshiftRequestContextFactory(final Settings settings, final RequestUtil
                 ConfigurationSettings.DEFAULT_OPENSHIFT_OPS_PROJECTS);
         this.kibanaPrefix = settings.get(ConfigurationSettings.KIBANA_CONFIG_INDEX_NAME, ConfigurationSettings.DEFAULT_USER_PROFILE_PREFIX);
         this.kibanaIndexMode = settings.get(ConfigurationSettings.OPENSHIFT_KIBANA_INDEX_MODE, UNIQUE);
-        if(!ArrayUtils.contains(new String [] {UNIQUE, SHARED_OPS}, kibanaIndexMode.toLowerCase())) {
+        if(!ArrayUtils.contains(new String [] {UNIQUE, SHARED_OPS, SHARED}, kibanaIndexMode.toLowerCase())) {
```
jcantrill: Hmm.. note to self there is an opportunity for this factory to rely on 'PluginSettings' so the call and initializations here are pushed into a common class.

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
```diff
@@ -169,10 +169,13 @@ private void addAliasToAllProjects(final OpenshiftRequestContext context, List<S
             if (filteredProjects.isEmpty()) {
                 filteredProjects.add(BLANK_PROJECT);
             } else {
-                String alias = UserRolesSyncStrategy.formatAllAlias(context.getUser());
-                LOGGER.debug("Adding indexes to alias '{}' for user '{}'", alias, context.getUser());
-                buildUserAlias(filteredProjects, projectPrefix, alias);
-                filteredProjects.add(alias);
+                // only add alias if kibana index mode is UNIQUE
+                if(context.getKibanaIndexMode() == UNIQUE) {
```
jcantrill: This is not quite correct.  An alias is added to simplify creating searches and visualizations across multiple indices.  I would expect maybe you would want a 'shared all' alias for non-ops users similar to the '.all' alias for ops

src/test/java/io/fabric8/elasticsearch/plugin/OpenshiftRequestContextFactoryTest.java
```diff
@@ -103,9 +103,14 @@ private void assertKibanaIndexIs(String mode) {
         switch (mode) {
         case "shared_ops":
             assertEquals("Exp. Kibana index mode to to be '.kibana'",
-                    ConfigurationSettings.DEFAULT_USER_PROFILE_PREFIX,
+                    ConfigurationSettings.DEFAULT_USER_PROFILE_PREFIX + "_ops",
```
jcantrill: This change will have ramifications introduced by earlier versions of the plugin for existing clusters where '.kibana' is for opts.  It will mean on upgrade their visualizations, searches, and dashboards would be lost to them.

src/main/java/io/fabric8/elasticsearch/plugin/KibanaIndexMode.java
FlorinPeter: the "shared" is enabling dashboard sharing for ops and non-ops (users) do you think we need a shared for both?

jcantrill: I don't know but should we ever (and i know thats not very agile to think of it know) but then shared for everyone would be.... 'global shared'?

FlorinPeter: ok you are right, but "shared_non_ops" is not the right key since, it enable the sharing for both, may we find another key ;)

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
FlorinPeter: This was a fix for the username alias, because it was shared among all users and it was a bit weird.

src/test/java/io/fabric8/elasticsearch/plugin/OpenshiftRequestContextFactoryTest.java
FlorinPeter: you are right so maybe it is a better way to let the index for ops to ".kibana" and add a new index ".kibana_user" or something like that, do you have a proposal?

src/main/java/io/fabric8/elasticsearch/plugin/OpenshiftRequestContextFactory.java
FlorinPeter: don't really know what you mean, I wanted to keep things identically

FlorinPeter:
@jcantrill I changed from SHARED to SHARED_NON_OPS may we find a better name but if not it is ok for me. I also added some tests. 
Thx again for your time and the great review.

src/main/java/io/fabric8/elasticsearch/plugin/KibanaIndexMode.java
jcantrill: I am open to another suggestion, any 'good' one escapes me at the moment

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
jcantrill: We are unable to use '.all' since alias names are global to the ES cluster.  We would just need to settle on something like '.all_non_ops' or something of the like.  I still am curious how this interplay will work given that users sharing the same kibana index will likely not share the same project indices.  Can you articulate a usecase or general motivation for this change?

FlorinPeter: The usecase or the request from our customers are currently the same when we rollout the EFK. They want to share searches and dashboards with each other, this is needed because our customers have a dedicated OCP for their users. I did not think about an alias to be honest and don't see currently the need, because as you mentioned, we maybe face the situation when not all users are share the same projects.

jcantrill: @portante What kind of issues due you think you will see with visualizations and dashboards if they are built using indices that may be only visible to some users?  Does this mean you will see the error bar at the top of Kibana for certain UI objects?

FlorinPeter: @jcantrill our QA team is currently testing the changes, will let you know if there are any findings.
If there is a saved search for example from for an index where the user don't have access then a security exception will be shown.
access_index_without_permission

portante: If somebody decides to use the `shared_non_ops` mode in a multi-tenant environment, does that even make sense?  But if somebody deploys their environment, where all users can see all non-operations related projects, then would be nice for sharing views.

portante:
I don't know what the right answer is, but I fear we are adding too many options to the mix here.  I wonder if we can take a step back and look at the use cases, and define a simple way to describe what use case we want to address rather than have multiple parameters that have to be tweaked to get the behavior we want.

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
FlorinPeter: @portante that's right, this option is not intended to be used on a multi-tenant environment. Our customer environments are configured in the way that users are seeing all non-operations related projects.

jcantrill:
Have you tried disabling Kibana rewrite [1].  The intention was for all users to share the same kibana index. Maybe that coupled with proper ACLs would produce your desired outcome

[1] https://github.com/fabric8io/openshift-elasticsearch-plugin/blob/master/src/main/java/io/fabric8/elasticsearch/plugin/ConfigurationSettings.java#L97

FlorinPeter:
@jcantrill the but then there are no indices configured and this means the user has to add his project manually, thats not really an option.

jcantrill:
@FlorinPeter That is not correct.  This setting was intended to turn off kibana transformations all together which means it will not convert your 'kibana' index request to something like 'kibana.usernamehash'.  TBH thought, I don't recall testing it in conjunction with some of the recent refactoring so I am not entirely certain it works as advertised.

FlorinPeter:
@jcantrill well, if I set openshift.kibana.rewrite.enabled: false then I get on my test environment "[security_exception] no permissions for indices:data/read/mget" while Kibana tries to access the .kibana index. With this option, if I understand it right, it should be possible to share .kibana with all users, this also includes ops. Did I understand it right?

So, if we combine "openshift.kibana.rewrite.enabled: false" with "openshift.kibana.index.mode: "shared_ops" then we maybe don't need my introduced option "openshift.kibana.index.mode: "shared_non_ops"

The result from my point of view, should be that OPs and non-Ops don't share the same .kibana index.

jcantrill:
Disabling rewrite, in theory, would cause all users to use the same kibana index.  As I said, I don't think the code properly accounts for how these settings should interplay with one another.  In any event, the default index for ops in this scenerio would be '.kibana'  so both ops and non ops users would be trying to use the same one.  The permissions are most likely not set correctly.

I think to achieve your desired behavior we still need an alternate index mode.  We also should consider removing the flag for the rewrite as I don't think it makes sense any longer with the concept of an index mode.

FlorinPeter:
@jcantrill sounds good to me, so what you propose, should we add another mode called "global" or do you have another idea? 

This means we will get 4 modes:
unique <- every ops and non-ops has his own index .kibana.userid
shared_ops <- ops are using .kibana and non-ops .kibana.userid
shared_no_ops <- ops are using .kibana and non-ops .kibana_non_ops
global <- ops and non-ops are using .kibana

Did I get it right?

jcantrill:
@FlorinPeter Correct, that would be 4 modes though we discussed only adding options as needed to simplify the testing matrix.  Please make the minimal changes you require to support your use case and we will leave the others to be done as the need arises.

FlorinPeter:
@jcantrill ok, so from my side this PR addresses the index mode "shared_no_ops" like we discussed.
Can you please review my latest changes and tell me if I should fix something or not it would be nice to see this merged ;)

jcantrill:
Prior to merge, please:

* rebase on master
* squash your commits

Please create a test PR https://github.com/openshift/origin-aggregated-logging that consumes a snapshot version of the artifact so we may confirm it additionally passes any regressions.

FlorinPeter:
@jcantrill I rebased on master and squashed my commits. I am not sure if changing this https://github.com/AppAgile/origin-aggregated-logging/commit/f32cf053f5a4035a58e01a600ea69d929c897902 will be enough, just wanted to ask before creating the PR.

pom.xml
```diff
@@ -26,7 +26,7 @@
 
   <groupId>io.fabric8.elasticsearch</groupId>
   <artifactId>openshift-elasticsearch-plugin</artifactId>
-  <version>2.4.4.13-SNAPSHOT</version>
+  <version>2.4.4.14-SNAPSHOT</version>
```
richm: I guess you added this for testing with origin-aggregated-logging?  This will have to be removed before being committed, or it will break the jenkins release job.

FlorinPeter: Well I merged my branch with the master and 2.4.4.14-SNAPSHOT is the latest version https://github.com/fabric8io/openshift-elasticsearch-plugin/blob/master/pom.xml#L29 maybe I missed something?

richm: Ah, I see - you didn't make that change - it was changed when you rebased on top of upstream?

FlorinPeter: Yes that's right

FlorinPeter:
@jcantrill looks good from my side https://github.com/openshift/origin-aggregated-logging/pull/638

pom.xml
jcantrill: Please revert this change.  One step of the 'release' workflow will update these values accordingly.

jcantrill:
@FlorinPeter please squash

FlorinPeter:
@jcantrill yes I am on it ;) give me a sec

FlorinPeter:
@jcantrill done

jcantrill:
Release job: https://fabric8-ci.fusesource.com/view/release%20builds/job/openshift-elasticsearch-plugin-release/60/

PR #97, opened 2017-08-28
fix empty indices for indices:data/read/field_stats

FlorinPeter:
This PR will fix the "[security_exception] no permissions for indices:data/read/msearch" by handling the empty indices when "data/read/field_stats" is called. A understandable error message is show to the user.

### DIFF
--- a/src/main/java/io/fabric8/elasticsearch/plugin/filter/FieldStatsResponseFilter.java
+++ b/src/main/java/io/fabric8/elasticsearch/plugin/filter/FieldStatsResponseFilter.java
@@ -21,6 +21,7 @@ import org.elasticsearch.action.ActionListener;
 import org.elasticsearch.action.ActionRequest;
 import org.elasticsearch.action.ActionResponse;
 import org.elasticsearch.action.fieldstats.FieldStatsRequest;
+import org.elasticsearch.action.fieldstats.FieldStatsResponse;
 import org.elasticsearch.action.support.ActionFilter;
 import org.elasticsearch.action.support.ActionFilterChain;
 import org.elasticsearch.common.inject.Inject;
@@ -72,6 +73,17 @@ public class FieldStatsResponseFilter implements ActionFilter {
             @SuppressWarnings("unchecked")
             @Override
             public void onResponse(ActionResponse response) {
+                if(INDICES_FIELD_STATS_READ_ACTION.equals(action) && response instanceof FieldStatsResponse) {
+                    if(((FieldStatsResponse)response).getIndicesMergedFieldStats().isEmpty()) {
+                        LOGGER.trace("Modifying the response to be {}", RestStatus.NO_CONTENT);
+                        Throwable err = new ElasticsearchException("The index returned an empty result. "
+                                + "You can use the Time Picker to change the time filter or select a higher time interval",
+                                RestStatus.NO_CONTENT);
+                        
+                        listener.onFailure(err);
+                        return;
+                    }
+                }
                 listener.onResponse(response);
             }
 


src/main/java/io/fabric8/elasticsearch/plugin/filter/FieldStatsResponseFilter.java
```diff
@@ -72,6 +73,17 @@ public void apply(final Task task, final String action, final ActionRequest requ
             @SuppressWarnings("unchecked")
             @Override
             public void onResponse(ActionResponse response) {
+                if(INDICES_FIELD_STATS_READ_ACTION.equals(action) && response instanceof FieldStatsResponse) {
```
jcantrill: There is a refactoring opportunity with this and onFailure since we appear to be checking the same thing

src/main/java/io/fabric8/elasticsearch/plugin/filter/FieldStatsResponseFilter.java
```diff
@@ -72,6 +73,17 @@ public void apply(final Task task, final String action, final ActionRequest requ
...
+                    if(((FieldStatsResponse)response).getIndicesMergedFieldStats().size() == 0) {
+                        LOGGER.trace("Modifying the response to be {}", RestStatus.NOT_FOUND);
```
jcantrill: Should this trace message be ' RestStatus.NO_CONTENT'

src/main/java/io/fabric8/elasticsearch/plugin/filter/FieldStatsResponseFilter.java
```diff
@@ -72,6 +73,17 @@ public void apply(final Task task, final String action, final ActionRequest requ
...
+                        Throwable err = new ElasticsearchException("The index is empty. This could mean data has not yet been collected.\n"
```
jcantrill: "...This could mean data has not yet been collected..." Is this message correct?  From an openshift perspective, I would expect this index to ONLY exist if the collector has created it.

src/main/java/io/fabric8/elasticsearch/plugin/filter/FieldStatsResponseFilter.java
```diff
@@ -72,6 +73,17 @@ public void apply(final Task task, final String action, final ActionRequest requ
...
+                                + "You can use the Time Picker to change the time filter or select a higher time interval",
+                                RestStatus.NO_CONTENT);
+                        
+                        listener.onFailure(err);
```
jcantrill: Should we be calling both 'onfailure' and 'onresponse' ?

src/main/java/io/fabric8/elasticsearch/plugin/filter/FieldStatsResponseFilter.java
FlorinPeter: yes I will fix it.

src/main/java/io/fabric8/elasticsearch/plugin/filter/FieldStatsResponseFilter.java
FlorinPeter: You are right, basically we need a message to explain the user that he need to choose a larger time or something like that, do you have a proposal?

src/main/java/io/fabric8/elasticsearch/plugin/filter/FieldStatsResponseFilter.java
FlorinPeter: well I was not sure if the onresponse is triggered when the onfailure was pervious called, but I can change it, I wanted to avoid if-else chaos ;)

src/main/java/io/fabric8/elasticsearch/plugin/filter/FieldStatsResponseFilter.java
jcantrill: Maybe: "There were no results found.  You may need to modify the time range by using the Time Picker to select a different interval."

src/main/java/io/fabric8/elasticsearch/plugin/filter/FieldStatsResponseFilter.java
jcantrill: Without actually testing I am unable to confirm or deny the code you have here.  I am just not familiar with it.

FlorinPeter:
@jcantrill thx for your review. 
I pushed some changes may you find time to check ;)

src/main/java/io/fabric8/elasticsearch/plugin/filter/FieldStatsResponseFilter.java
```diff
@@ -72,7 +73,20 @@ public void apply(final Task task, final String action, final ActionRequest requ
             @SuppressWarnings("unchecked")
             @Override
             public void onResponse(ActionResponse response) {
-                listener.onResponse(response);
+                if(INDICES_FIELD_STATS_READ_ACTION.equals(action) && response instanceof FieldStatsResponse) {
+                    if(((FieldStatsResponse)response).getIndicesMergedFieldStats().size() == 0) {
```
jcantrill: I should have caught this earlier but can you say '.getIndicesMergedFieldStats().isEmpty()'

src/main/java/io/fabric8/elasticsearch/plugin/filter/FieldStatsResponseFilter.java
```diff
@@ -72,7 +73,20 @@ public void apply(final Task task, final String action, final ActionRequest requ
...
+                        LOGGER.trace("Modifying the response to be {}", RestStatus.NO_CONTENT);
+                        Throwable err = new ElasticsearchException("The index returned an empty result."
+                                + "You can use the Time Picker to change the time filter or select a higher time interval",
+                                RestStatus.NO_CONTENT);
+                        
+                        listener.onFailure(err);
+                    } else {
```
jcantrill: Probably instead of the nested if/else, simply utilize an earlier return after 'onfailure'
```
if(FieldStatsResponse){
...
onFailure(err)
return
}
onResponse()
```

FlorinPeter:
@jcantrill ok, we have a good flow ;)

jcantrill:
Can you squash these changes for them to be merged

FlorinPeter:
@jcantrill yes, I will do it tomorrow and give you a ping, thx.

FlorinPeter:
@jcantrill sorry for the delay, I squashed the changes.

PR #100, opened 2017-09-06
bug 1476062. Only seed documents when they are missing

jcantrill:
This PR

* Modifies seeding code to only create index-patterns and alias when they do not already exist
* Requires an additional permission for the role to allow write to the kibana index

### DIFF
--- a/src/main/java/io/fabric8/elasticsearch/plugin/PluginClient.java
+++ b/src/main/java/io/fabric8/elasticsearch/plugin/PluginClient.java
@@ -16,19 +16,28 @@
 
 package io.fabric8.elasticsearch.plugin;
 
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 import org.elasticsearch.action.ActionRequestBuilder;
 import org.elasticsearch.action.admin.indices.alias.IndicesAliasesRequestBuilder;
 import org.elasticsearch.action.admin.indices.alias.IndicesAliasesResponse;
+import org.elasticsearch.action.admin.indices.alias.get.GetAliasesRequestBuilder;
+import org.elasticsearch.action.admin.indices.alias.get.GetAliasesResponse;
 import org.elasticsearch.action.admin.indices.exists.indices.IndicesExistsRequestBuilder;
 import org.elasticsearch.action.admin.indices.exists.indices.IndicesExistsResponse;
+import org.elasticsearch.action.get.GetRequestBuilder;
+import org.elasticsearch.action.get.GetResponse;
+import org.elasticsearch.action.index.IndexRequestBuilder;
+import org.elasticsearch.action.index.IndexResponse;
 import org.elasticsearch.client.Client;
 import org.elasticsearch.common.inject.Inject;
 import org.elasticsearch.common.logging.ESLogger;
 import org.elasticsearch.common.logging.Loggers;
 
 import com.floragunn.searchguard.support.ConfigConstants;
+import com.google.common.collect.UnmodifiableIterator;
 
 /**
  * Facade to the ES client to simplify calls
@@ -43,6 +52,14 @@ public class PluginClient {
     public PluginClient(Client client) {
         this.client = client;
     }
+    
+    public IndexResponse createDocument(String index, String type, String id, String source) {
+        LOGGER.trace("create document: '{}/{}/{}' source: '{}'", index, type, id, source);
+        IndexRequestBuilder builder = client.prepareIndex(index, type, id).setSource(source);
+        addCommonHeaders(builder);
+        IndexResponse response = builder.get();
+        return response;
+    }
 
     public boolean indexExists(final String index) {
         LOGGER.trace("Checking for existance of index '{}'", index);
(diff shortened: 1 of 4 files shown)


jcantrill:
cc @portante

src/main/java/io/fabric8/elasticsearch/plugin/PluginClient.java
```diff
@@ -53,6 +70,40 @@ public boolean indexExists(final String index) {
         LOGGER.trace("Index '{}' exists? {}", index, exists);
         return exists;
     }
+    
+    public boolean documentExists(final String index, final String type, final String id) {
+        LOGGER.trace("Checking for existance of document: '{}/{}/{}'", index, type, id);
```
richm: existence

src/main/java/io/fabric8/elasticsearch/plugin/PluginClient.java
```diff
@@ -53,6 +70,40 @@ public boolean indexExists(final String index) {
...
+     * Retrieve the set of indices for a given alias
+     * 
+     * @param alias The alias to lookup
+     * @return The set of indices to the given alias
+     */
+    public Set<String> getIndicesForAlias(String alias){
+        LOGGER.trace("Retrieving indices for alias '{}'", alias);
+        GetAliasesRequestBuilder builder = this.client.admin().indices().prepareGetAliases(alias);
+        addCommonHeaders(builder);
+        GetAliasesResponse response = builder.get();
```
richm: What happens if `alias` doesn't exist?  Will `builder.get()` return a `404`?  Will it throw an exception?  Will `response.getAliases().keysIt()` return `null`?  an "empty" iterator?

src/main/java/io/fabric8/elasticsearch/util/IndexUtil.java
```diff
@@ -0,0 +1,46 @@
...
+     * 
+     * @param value     the value to use in place of the date
+     * @param indices   the set of indices to process
+     * @return  A set of the indices with modified values
+     */
+    public Set<String> replaceDateSuffix(final String value, final Set<String> indices){
+        Set<String> modified = new HashSet<>();
+        final String sub = "." + value;
+        for (String index : indices) {
+            modified.add(index.replaceFirst("\\.\\d\\d\\d\\d\\.\\d\\d.\\d\\d", sub));
```
richm: can you anchor this match to the end of the string e.g.
```
modified.add(index.replaceFirst("\\.\\d\\d\\d\\d\\.\\d\\d.\\d\\d$", sub));
```
or something like that?

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
```diff
@@ -269,9 +269,14 @@ private void buildUserAlias(List<String> projects, String projectPrefix, String
             if (projects.isEmpty()) {
                 return;
             }
+            Set<String> aliasedIndicies = pluginClient.getIndicesForAlias(alias);
+            aliasedIndicies = new IndexUtil().replaceDateSuffix("*", aliasedIndicies);
             Map<String, String> patternAlias = new HashMap<>(projects.size());
             for (String project : projects) {
-                patternAlias.put(getIndexPattern(project, projectPrefix), alias);
+                String indexPattern = getIndexPattern(project, projectPrefix);
+                if(!aliasedIndicies.contains(indexPattern)) {
```
richm: do the `indexPattern, alias` pairs in `aliasedIndices` get added to `patternAlias` somewhere else, or are they not needed any more?

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
```diff
@@ -318,11 +323,15 @@ private String getDefaultIndex(OpenshiftRequestContext context, Client esClient,
...
 
     private void create(String kibanaIndex, List<String> projects, boolean setDefault, Client esClient,
-            String kibanaVersion, String projectPrefix) {
+            String kibanaVersion, String projectPrefix, Set<String> indexPatterns) {
         boolean defaultSet = !setDefault;
-
+        LOGGER.trace("Creating index-patterns for projects: '{}'", projects);
         for (String project : projects) {
-            createIndex(kibanaIndex, project, esClient, projectPrefix);
+            if(indexPatterns.contains(project)) { //no need to update
```
richm: What happens when we upgrade, and the new es image has new index patterns?  Do we have to manually delete the old index patterns first? I guess the real problem is when the user has created their own index patterns, but we need to add new ones during an upgrade.

richm:
Can we have a test to verify that an existing index pattern is not replaced?

src/main/java/io/fabric8/elasticsearch/plugin/PluginClient.java
jcantrill: An empty iterator is returned.

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
jcantrill: They are not needed any longer and are discarded.

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
jcantrill: This is being driven by observations from @portante.  When the index patterns change, the only way to update them would be to delete the kibana index or the document in question.  The sole purpose of this PR is to avoid the issue seen by users where they build visualizations and dashboards based upon index mappings and then the mappings change.  The only solution I can thing of  would be to create an admin utility to allow updating mappings for a given user.  It might be easier to just delete there index.

jcantrill:
@richm I'm still working on how to best run these tests consistently with our CI but was able to prove that:
* Before this change and when write permissions are missing the test fails
* After the change with write permissions the test succeeds.

Note: The steps are subtly different but it is only the verbiage and an initialization step

Before fix:
```
Feature: An user's kibana index is seeded with index mappings for
  the indices that are visible to them.

  Scenario: A user's kibana index is only seeded once     # features/elasticsearch.feature:7
    Given a user with the role of operations              # features/step_definitions/policy_steps.rb:1
    And logs were collected for the operations project    # features/step_definitions/elasticsearch_steps.rb:1
    And they log into Kibana                              # features/step_definitions/kibana_steps.rb:1
    And refresh their index field list                    # features/step_definitions/kibana_steps.rb:25
      Failed to write when refreshing field mappings. UI says 155 but index says 251 (RuntimeError)
      ./features/step_definitions/kibana_steps.rb:69:in `/^refresh their index field list$/'
      ./features/support/hooks.rb:15:in `call'
      ./features/support/hooks.rb:15:in `block in <top (required)>'
      features/elasticsearch.feature:14:in `And refresh their index field list'
    When they log into Kibana again                       # features/step_definitions/kibana_steps.rb:15
    Then the index mapping fields should remain unchanged # features/step_definitions/kibana_steps.rb:72

Failing Scenarios:
cucumber features/elasticsearch.feature:7 # Scenario: A user's kibana index is only seeded once

1 scenario (1 failed)
6 steps (1 failed, 2 skipped, 3 passed)
0m17.303s
```
After fix:
```
@elasticsearch
Feature: An user's kibana index is seeded with index mappings for
  the indices that are visible to them.

  Scenario: A user's kibana index is only seeded once     # features/elasticsearch.feature:7
    Given a user with the role of operations              # features/step_definitions/policy_steps.rb:1
    And the user has never used Kibana                    # features/step_definitions/kibana_steps.rb:15
    And logs were collected for the operations project    # features/step_definitions/elasticsearch_steps.rb:1
    And they log into Kibana                              # features/step_definitions/kibana_steps.rb:1
    And refresh operations index pattern field list       # features/step_definitions/kibana_steps.rb:34
    When they log into Kibana again                       # features/step_definitions/kibana_steps.rb:24
    Then the index mapping fields should remain unchanged # features/step_definitions/kibana_steps.rb:74

1 scenario (1 passed)
7 steps (7 passed)
0m20.763s
````
Does this satisfy your test request?

jcantrill:
cc @portante

PR #101, opened 2017-09-14
bug 1491227. Modify request if user has back slash

jcantrill:
This PR fixes BZ1491227:
* Modifying the request header to convert backslash to forward
* Fixes role and role mapping generation to handle slashes

### DIFF
--- a/src/main/java/io/fabric8/elasticsearch/plugin/OpenshiftRequestContextFactory.java
+++ b/src/main/java/io/fabric8/elasticsearch/plugin/OpenshiftRequestContextFactory.java
@@ -83,6 +83,10 @@ public class OpenshiftRequestContextFactory  {
         Set<String> projects = new HashSet<>();
         boolean isClusterAdmin = false;
         String user = utils.getUser(request);
+        if(user.contains("\\")){
+            user = user.replaceAll("\\\\", "/");
+            utils.setUser(request, user);
+        }
         String token = utils.getBearerToken(request);
         if (StringUtils.isNotBlank(user) && StringUtils.isNotBlank(token)){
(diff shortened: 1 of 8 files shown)


jcantrill:
cc @portante

src/main/java/io/fabric8/elasticsearch/plugin/OpenshiftRequestContextFactory.java
```diff
@@ -83,6 +83,10 @@ public OpenshiftRequestContext create(final RestRequest request, final UserProje
         Set<String> projects = new HashSet<>();
         boolean isClusterAdmin = false;
         String user = utils.getUser(request);
+        if(user.contains("\\")){
+            user = user.replaceAll("\\\\", "/");
```
ewolinetz: We really need to escape this much? Is there a java util we can use to escape for us so this looks cleaner?

src/main/java/io/fabric8/elasticsearch/plugin/acl/BaseRolesSyncStrategy.java
```diff
@@ -71,7 +71,7 @@ protected String formatKibanaRoleName(UserProjectCache cache, String username, S
     }
     
     public static String formatUserRoleName(String username) {
-        return String.format("%s_%s", USER_PREFIX, username.replaceAll("[\\.@]", "_"));
+        return String.format("%s_%s", USER_PREFIX, username.replaceAll("[\\\\.@/]", "_"));
```
ewolinetz: Similar question about escaping so much

ewolinetz:
Quick questions, otherwise LGTM

PR #108, opened 2017-10-10
1456584 ldap names

jcantrill:
This PR:
* fixes the case where usrnames provided from LDAP causes errors
* removes the 'all' alias for non-ops users as they should be able to query across their projects using 'project.*'

cc @portante


jcantrill:
[merge]

PR #109, opened 2017-10-13
bug 1500086. Return Unauthorized if username does not match bearer token

jcantrill:
This PR:
* Closes the gap where a user can auth creating a role and then a second user can use the same username without providing a token

cc @simo5 @pweil-


src/it/java/io/fabric8/elasticsearch/ElasticsearchIntegrationTest.java
```diff
@@ -304,16 +299,29 @@ protected void givenDocumentIsIndexed(String index, String type, String id, XCon
...
+        apiServer.expect().get().withPath("/oapi/v1/users/~").andReturn(200,response).once();
+    }
+
+    protected void givenTokenIsUnAuthorizedForUser() {
+        apiServer.expect().get().withPath("/oapi/v1/users/~").andReturn(401,"Unauthorized").once();
+    }
+    
+    private void expSubjectAcccessReviewToBe(boolean value, String user, String token) {
         testContext.put(USERNAME, user);
+        testContext.put(TOKEN, token);
```
portante: Where is `testContext` used here?

src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java
```diff
@@ -73,6 +81,42 @@ public boolean isOperationsUser(RestRequest request) {
...
+        final String token = getBearerToken(request);
+        ConfigBuilder builder = new ConfigBuilder().withOauthToken(token);
+        try (DefaultOpenShiftClient osClient = new DefaultOpenShiftClient(builder.build())) {
+            LOGGER.debug("Verifying user {} matches the given token.", user);
+            Request okRequest = new Request.Builder()
+                .addHeader(AUTHORIZATION_HEADER, "Bearer " + token)
+                .url(osClient.getMasterUrl() + "oapi/v1/users/~") 
+                .build();
+            String username = null;
+            try (Response response = osClient.getHttpClient().newCall(okRequest).execute()){
```
portante: Shouldn't this assignment be in the `try` block?

src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java
```diff
@@ -73,6 +81,42 @@ public boolean isOperationsUser(RestRequest request) {
...
+                if(response.code() != RestStatus.OK.getStatus()) {
+                    throw new ElasticsearchSecurityException("The user is unauthorized", RestStatus.UNAUTHORIZED);
+                }
+                String body = response.body().string();
+                if (LOGGER.isDebugEnabled()) {
+                    LOGGER.debug("Response: {}", new String(body));
```
portante: Isn't `body` already a `String`?

src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java
```diff
@@ -73,6 +81,42 @@ public boolean isOperationsUser(RestRequest request) {
...
+        final String user = getUser(request);
+        final String token = getBearerToken(request);
+        ConfigBuilder builder = new ConfigBuilder().withOauthToken(token);
+        try (DefaultOpenShiftClient osClient = new DefaultOpenShiftClient(builder.build())) {
+            LOGGER.debug("Verifying user {} matches the given token.", user);
+            Request okRequest = new Request.Builder()
+                .addHeader(AUTHORIZATION_HEADER, "Bearer " + token)
+                .url(osClient.getMasterUrl() + "oapi/v1/users/~") 
+                .build();
+            String username = null;
```
portante: Shouldn't `username` be declared outside the `try` block scope so that it will be in-scope for line 113?

src/it/java/io/fabric8/elasticsearch/ElasticsearchIntegrationTest.java
jcantrill: someplace to stuff values for reuse in subsequent steps.  Its really to make the test read nicer without the need to return values from the statments

src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java
jcantrill: No.  On line 113 there is an explicit comparison that is outside the try block.

src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java
jcantrill: Not its a responsebody type or something specific to the library impl

portante: But on line 101 `body` appears to be defined as a `String`.  So why do we need to do `new String(body)` now.

src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java
```diff
@@ -73,6 +81,42 @@ public boolean isOperationsUser(RestRequest request) {
...
+                    username = (String) ((Map)userResponse.get("metadata")).get("name");
+                }
+            }catch (Exception e) {
+                LOGGER.debug("Exception trying to assertUser '{}'", e, user);
+                throw e;
+            }
+            if(!user.equals(username)) {
+                String message = String.format("The username '%s' does not own the token provided with the request.", username);
+                LOGGER.debug(message);
+                throw new ElasticsearchSecurityException(message, RestStatus.FORBIDDEN);
```
pweil-: does this message appear anywhere for the user?  If so it may be an info leak.  @simo5, thoughts?

src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java
```diff
@@ -73,6 +81,42 @@ public boolean isOperationsUser(RestRequest request) {
...
+        try (DefaultOpenShiftClient osClient = new DefaultOpenShiftClient(builder.build())) {
+            LOGGER.debug("Verifying user {} matches the given token.", user);
+            Request okRequest = new Request.Builder()
+                .addHeader(AUTHORIZATION_HEADER, "Bearer " + token)
+                .url(osClient.getMasterUrl() + "oapi/v1/users/~") 
+                .build();
+            String username = null;
+            try (Response response = osClient.getHttpClient().newCall(okRequest).execute()){
+                if(response.code() != RestStatus.OK.getStatus()) {
+                    throw new ElasticsearchSecurityException("The user is unauthorized", RestStatus.UNAUTHORIZED);
```
pweil-: always ok to hide the response code here with unauthorized?  Maybe add a debug for the real response code just in case.

simo5:
Do you really have to use the access token ?
I wonder if avoiding it and always using oauth-proxy (with a shared secret between oauth-proxy and you to secure communication) wouldn't be a better solution long term.

src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java
jcantrill: It may appear in the response to the call, but if we are not supplying the token in the response i that an issue? Also, a proxy or someone manually curling would have this info already?  Is that an issue?

src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java
jcantrill: I wanted to throw an exception with the returned code from the other call but the only class that take s a code was "SecurityException"

src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java
pweil-: generally unauthorized is simply unauthorized without anymore info

jcantrill:
@simo5 We need to know:

* the project that are visible for a given user
* Their role as a cluster admin (or not)

There is no way for us to get this information without the users token unless something has come about since the original writing of this plugin

src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java
jcantrill: This is generally true I believe.  Looking at the dependent code, however, this is a basically a builder pattern that is deferred until execute.  I'm going to assume the assignment is fine.

src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java
jcantrill: fixed and changed to unauthorized which is supposed to provide an auth url per the spec.  Im going to forgo adding that as: 1. not sure what to set it to, 2. its not a true web service and you should know where to get your token.

src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java
jcantrill: Removed message and modified to have debug log message with code

jcantrill:
@richm Updated the PR with variance of your changes.

PR #112, opened 2017-11-28
Reenable IT ci testing

ewolinetz:
Reenabled integration testing using the failsafe plugin.
We run this with the goal `verify` and profile `release`


ewolinetz:
IT tests are currently not passing locally... need to gen some keys that were around before

ewolinetz:
@jcantrill @richm PTAL
Also, do we want this to run all the time with the other surefire tests or just with `release:verify` ?

richm:
I would ideally like to have these integration tests run against every PR and block merging

ewolinetz:
I'm not able to figure out why the build test is failing. running tasks `clean compile` locally passes for me

ewolinetz:
[retest]

ewolinetz:
Fixed compiler errors, but integration tests are failing now...

src/test/java/io/fabric8/elasticsearch/integrationtest/ProxyUsernameIntegrationTest.java
```diff
@@ -0,0 +1,65 @@
...
+        assertThatResponseIsForbidden();
+    }
+    
+    @Test
+    public void testUsersWithSimpleNameCreatesProfileIndex() throws Exception {
+        startES();
+
+        String [] users = {"foo@email.com", 
+            "CN=jdoe,OU=DL IT,OU=User Accounts,DC=example,DC=com", 
+            "test/username",
```
ewolinetz: this used to be "test\\\\username" before and "CN=Lastname\\\\\\\\ , Firstname...." however it seems SG is converting "\\" to "/" internally... not sure if thats something we should test/fix though?

@jcantrill @richm

jcantrill: It has to be tested with the backward slashes as that is directly related to issues we have seen in the field.  We modify that value [1] .  It may be mitigated with the latest code where hashes the values to generate the roles, but they should remain the same because the proxy header will have to match the entry in the rolesmapping.yml

[1] https://github.com/fabric8io/openshift-elasticsearch-plugin/blob/master/src/main/java/io/fabric8/elasticsearch/plugin/OpenshiftRequestContextFactory.java#L91

ewolinetz: Got it, I needed to do some more escaping for my mock response...

jcantrill:
closing in favor of #152 as I dont see us fixing for 2.4.4 line

PR #115, opened 2017-12-04
fix 1519705. All alias shows now indices

jcantrill:
This PR fixes 1519705:

* Alias .operations if operations user and .operation indices exist

```
[2017-12-04 21:02:58,033][ERROR][io.fabric8.elasticsearch.plugin.kibana.KibanaSeed] Error executing Alias request
[.operations.*] IndexNotFoundException[no such index]

```

### DIFF
--- a/src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
+++ b/src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
@@ -99,6 +99,9 @@ public class KibanaSeed implements ConfigurationSettings {
         LOGGER.debug("Found '{}' Index patterns for user", indexPatterns.size());
 
         Set<String> projects = new HashSet<>(context.getProjects());
+        if(context.isOperationsUser()) {
+            projects.add(OPERATIONS_PROJECT);
+        }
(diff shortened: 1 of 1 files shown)


jcantrill:
I only manually tested because without integration tests this is very difficult to set up

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
```diff
@@ -99,6 +99,9 @@ public void setDashboards(final OpenshiftRequestContext context, Client client,
         LOGGER.debug("Found '{}' Index patterns for user", indexPatterns.size());
 
         Set<String> projects = new HashSet<>(context.getProjects());
+        if(context.isOperationsUser()) {
+            projects.add(OPERATIONS_PROJECT);
+        }
```
richm: Ah, so move it here, so it will get filtered out of the list if there is no `.operations.*` indices?  Ok

jcantrill: Correct.  Assume there are indices and then when we check for their existence it will be filtered out of the set

PR #120, opened 2018-02-07
Allow cluster admins view logs from no longer existing projects

josefkarasek:


josefkarasek:
@jcantrill @ewolinetz This might be naive solution, please review/advice.

Steps I used for testing:
* Create project and deploy some app that writes logs - both, project owner and cluster admin can view logs
* Delete project - only cluster admin can view logs
* Delete index from es - no logs in es, no index mapping for project in .kibana index

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
```diff
@@ -96,11 +96,18 @@ public void setDashboards(final OpenshiftRequestContext context, Client client,
         // compare results to projects; handle any deltas (create, delete?)
 
         Set<String> indexPatterns = getProjectNamesFromIndexes(context, client, projectPrefix);
-        LOGGER.debug("Found '{}' Index patterns for user", indexPatterns.size());
+        LOGGER.debug("Found '{}' Index patterns for user {}", indexPatterns.size(), indexPatterns);
```
jcantrill: @richm  @ewolinetz any thoughts if this is a security concern dumping visible indices to the logs?

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
```diff
@@ -96,11 +96,18 @@ public void setDashboards(final OpenshiftRequestContext context, Client client,
...
 
         Set<String> projects = new HashSet<>(context.getProjects());
         if(context.isOperationsUser()) {
             projects.add(OPERATIONS_PROJECT);
+            for (String pattern : indexPatterns) {
+                // Project might no longer exist, but the data can still be retrieved from elasticsearch
+                if (!projects.contains(pattern) && !pattern.equals(ADMIN_ALIAS_NAME)) {
```
jcantrill: I don't believe alias are returned as part of the listed indices which I would think its then not necessary to compare against those

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
```diff
@@ -136,8 +143,20 @@ public void setDashboards(final OpenshiftRequestContext context, Client client,
...
-            // cull any that are in ES but not in OS (remaining in indexPatterns)
-            remove(context.getKibanaIndex(), indexPatterns, client, projectPrefix);
+            if (!context.isOperationsUser()) {
+                // cull any that are in ES but not in OS (remaining in indexPatterns)
+                remove(context.getKibanaIndex(), indexPatterns, client, projectPrefix);
+            } else {
+                // delete index patterns only if the data has been deleted
+                Set<String> patternsWithoutData = new HashSet<>(indexPatterns.size());
+                for (String pattern : indexPatterns) {
+                    if (!pluginClient.indexExists(getIndexPattern(pattern, projectPrefix))) {
```
jcantrill: Does this work for wildcards?

ewolinetz:
How is this different than the current logic where a cluster admin can view old logs of deleted namespaces within the `.all` alias?

jcantrill:
@ewolinetz it could be our logic currently removes the alias from the index.  I have not reviewed in depth to be certain

jcantrill:
Closing for now.  We should be able to do this using the '.all' alias or by the fact admins can see all indices

PR #121, opened 2018-02-13
add use of openssl for keystore

jcantrill:
Add openssl dependent libraries


richm:
Why do we need this?  We can't use the native Java SSL implementation because . . .?

jcantrill:
Only reason I can offer:

```
Using OpenSSL will result in better performance and better support for strong and modern cipher suites when compared JCE
```

ewolinetz:
Is it that simple for enabling ssl for SG?
LGTM

PR #123, opened 2018-03-15
impl single SG index

jcantrill:
This PR allows our plugin to dynamically update a single SG instance:

* Uses several tries with backoff to update ACL documents based on version
* Runs expiration thread on master nodes only and overwrites ACL document to removed expires
* Ignores roles and mappings where expiration value is not defined

cc @portante


jcantrill:
@portante @ewolinetz I'm missing something on the expiration thread.  I see it start in the logs, I see the actual thread when I connect via a debugger, I dont ever see it execute.  Are my time values wrong?

src/main/java/io/fabric8/elasticsearch/plugin/acl/ACLDocumentManager.java
```diff
@@ -0,0 +1,283 @@
...
+
+        private OpenshiftRequestContext context;
+
+        public SyncFromContextOperation(OpenshiftRequestContext context) {
+            this.context = context;
+        }
+        
+        @Override
+        public void syncDocuments(Collection<SearchGuardACLDocument> docs) {
+            LOGGER.debug("Syncing from context to ACL...");
```
ewolinetz: isDebugEnabled

src/main/java/io/fabric8/elasticsearch/plugin/acl/ACLDocumentManager.java
```diff
@@ -0,0 +1,283 @@
...
+
+        private long now;
+
+        public ExpireOperation(long currentTimeMillis) {
+            this.now = currentTimeMillis;
+        }
+
+        @Override
+        public void syncDocuments(Collection<SearchGuardACLDocument> docs) {
+            LOGGER.debug("Expiring ACLs older then {}", now);
```
ewolinetz: isDebugEnabled

src/main/java/io/fabric8/elasticsearch/plugin/acl/ACLDocumentManager.java
```diff
@@ -0,0 +1,283 @@
...
+            for (SearchGuardACLDocument doc : docs) {
+                if(ConfigurationSettings.SEARCHGUARD_MAPPING_TYPE.equals(doc.getType())){
+                    SearchGuardRolesMapping mappings = (SearchGuardRolesMapping) doc;
+                    for (RolesMapping mapping : mappings) {
+                        //assume if the value is there its intentional
+                        if(mapping.getExpire() != null && mapping.getExpire().longValue() < now) {
```
ewolinetz: does `.getExpire()` return the time when this will expire? if so, should this be `>=` ?

src/main/java/io/fabric8/elasticsearch/plugin/acl/ACLDocumentManager.java
```diff
@@ -0,0 +1,283 @@
...
+                }
+                Thread.sleep(n * 1000);
+            } catch (InterruptedException e) {
+                LOGGER.error("There was an error while trying the sleep the syncACL operation", e);
+            }
+        }
+        return false;
+    }
+    
+    public boolean trySyncAcl(ACLDocumentOperation operation) {
```
ewolinetz: could we make this method sychronized instead of using locks?

src/main/java/io/fabric8/elasticsearch/plugin/acl/ACLDocumentManager.java
```diff
@@ -0,0 +1,283 @@
...
+            LOGGER.error("Exception while syncing ACL with cache", e);
+        } finally {
+            lock.unlock();
+        }
+        return false;
+    }
+    
+    
+    
+    private Collection<SearchGuardACLDocument> loadAcls() throws Exception {
```
ewolinetz: could we make this method sychronized instead of using locks surrounding when we call this?

src/main/java/io/fabric8/elasticsearch/plugin/acl/ACLDocumentManager.java
```diff
@@ -0,0 +1,283 @@
...
+                    LOGGER.debug("Successfully reloaded config with '{}' nodes", cur.getNodes().length);
+                }else {
+                    LOGGER.warn("Failed to reloaded configs", cur.getNodes().length);
+                }
+            }catch(Exception e) {
+                LOGGER.error("Unable to notify of an ACL config update", e);
+            }
+            return true;
+        } else {
+            LOGGER.debug("Unable to write ACL {}", response.buildFailureMessage());
```
ewolinetz: isDebugEnabled

src/main/java/io/fabric8/elasticsearch/plugin/acl/ACLDocumentManager.java
```diff
@@ -0,0 +1,283 @@
...
+    }
+    
+    private boolean isSuccessfulWrite(BulkResponse response) {
+        if(!response.hasFailures()) {
+            ConfigUpdateRequest confRequest = new ConfigUpdateRequest(SEARCHGUARD_INITIAL_CONFIGS);
+            confRequest.putHeader(ConfigConstants.SG_CONF_REQUEST_HEADER, "true");
+            try {
+                ConfigUpdateResponse cur = this.client.execute(ConfigUpdateAction.INSTANCE, confRequest).actionGet();
+                if (cur.getNodes().length > 0) {
+                    LOGGER.debug("Successfully reloaded config with '{}' nodes", cur.getNodes().length);
```
ewolinetz: isDebugEnabled

src/main/java/io/fabric8/elasticsearch/plugin/acl/ACLDocumentManager.java
```diff
@@ -0,0 +1,283 @@
...
+    }
+    
+    private BulkResponse writeAcl(ACLDocumentOperation operation, Collection<SearchGuardACLDocument> docs) throws Exception {
+        BulkRequestBuilder builder = client.prepareBulk().setRefresh(true);
+        BulkRequest request = operation.buildRequest(this.client, builder, docs);
+        request.putHeader(ConfigConstants.SG_CONF_REQUEST_HEADER, "true");
+        return this.client.bulk(request).actionGet();
+    }
+    
+    private boolean isSuccessfulWrite(BulkResponse response) {
```
ewolinetz: same question here as loadAcls regarding being synchronized

src/main/java/io/fabric8/elasticsearch/plugin/OpenShiftElasticSearchService.java
```diff
@@ -31,82 +33,97 @@
...
-        this.cache = cache;
         restController.registerFilter(aclFilter);
+        this.aclManager = aclManager;
+        clusterService.add(this);
     }
-
+    
     @Override
     protected void doStart() throws ElasticsearchException {
+        logger.debug("Started");
```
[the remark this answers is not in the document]
jcantrill: The compiler will optimize this string and the logger, internally, will treat this is a noop unless debug is enabled.

src/main/java/io/fabric8/elasticsearch/plugin/acl/ACLDocumentManager.java
jcantrill: Same here

src/main/java/io/fabric8/elasticsearch/plugin/acl/ACLDocumentManager.java
jcantrill: getExpire returns a java.util.Long being the millis since epoch.  It will only remove the mappings when the check time 'now' is 'older' then the expire time which should be '<'

src/main/java/io/fabric8/elasticsearch/plugin/acl/ACLDocumentManager.java
jcantrill: Sync will lock the entire object.  By the way we have implemented lock it is possible we have fundamentally already done that.

richm:
Can you explain the strategy to ensure only 1 shared searchguard index per cluster?  How do you ensure that updates to the index are atomic?

jcantrill:
Closing for now as unlikely to release for the 2.4.x line

PR #134, opened 2018-05-18
impl single SG index

jcantrill:
This PR modifies ACL generation:

* utilizes a single SG index
* attempts to reconcile based on version conflicts
* removes all plugin caching
* expires generated acls based on some expire period

This is the 5.x release of #123


jcantrill:
cc @t0ffel

jcantrill:
cc @portante

src/main/java/io/fabric8/elasticsearch/plugin/OpenShiftElasticSearchService.java
```diff
@@ -23,82 +23,93 @@
...
 
             Runnable expire = new Runnable() {
                 @Override
                 public void run() {
-                    cache.expire();
+                    aclManager.expire();
                 }
             };
-            this.scheduledFuture = this.scheduler.scheduleWithFixedDelay(expire, 5, 60, TimeUnit.SECONDS);
+            this.scheduledFuture = this.scheduler.scheduleWithFixedDelay(expire, 1000 * 60, settings.getACLExpiresInMillis(), TimeUnit.MILLISECONDS);
```
richm: what does this do?

src/main/java/io/fabric8/elasticsearch/plugin/acl/ACLDocumentManager.java
```diff
@@ -0,0 +1,287 @@
...
+    
+    private boolean syncAcl(ACLDocumentOperation operation) {
+        //try up to 30 seconds and then continue
+        for (int n : new int [] {1 , 1 , 2 , 3 , 5 , 8}) {
+            if(trySyncAcl(operation)) {
+                return true;
+            }
+            try {
+                if(LOGGER.isTraceEnabled()) {
+                    LOGGER.trace("Sleeping for {}(s)", n * 1000);
```
richm: n is seconds?  So just `n` not `n * 1000`

src/main/java/io/fabric8/elasticsearch/plugin/OpenShiftElasticSearchService.java
jcantrill: This schedules a repeating task to expire ACLs based on the config value.  The process begins 1 minute after loading.

jcantrill:
In response to @richm 
```
Can you explain the strategy to ensure only 1 shared searchguard index per cluster? 
How do you ensure that updates to the index are atomic?
```
The algorithm is based upon the recommendations of Elastic and uses optimistic concurrency and document versioning.  We load, grab version, and attempt to update which will merge our changes into the existing document and then bump the version.  If this fails, it backs off, reloads, and tries again.  Additionally, every X seconds we expire entries which is a full document replace since 'update' is only additive; ES makes no effort to remove sections.

Basically, each node will make a single change and continue to try to submit that changes until it is accepted.  @portante and I determined removing any notion of caching was acceptable given this is only handing requests with bearer tokens which should be minimal.  Conceptually there probably is a case to be made for some sort of DOS if servicing requests if their are high rates using bearer tokens.

src/main/java/io/fabric8/elasticsearch/plugin/OpenShiftElasticSearchService.java
```diff
@@ -23,82 +23,93 @@
...
     protected void doStart() throws ElasticsearchException {
+        LOGGER.debug("Started");
+    }
 
-        boolean dynamicEnabled = settings.getAsBoolean(OPENSHIFT_DYNAMIC_ENABLED_FLAG,
-                OPENSHIFT_DYNAMIC_ENABLED_DEFAULT);
-        logger.debug("Starting with Dynamic ACL feature enabled: {}", dynamicEnabled);
-
+    @Override
+    public void onMaster() {
```
ewolinetz: when does `onMaster` and `offMaster` trigger?

ewolinetz:
@jcantrill how do we ensure that our pre-populated ACL definitions don't expire. Is it simply because they'll have a `null` `expires` field?

src/main/java/io/fabric8/elasticsearch/plugin/OpenShiftElasticSearchService.java
jcantrill: It's a callback from Elasticsearch when a node is elected master or drops out

jcantrill:
@ewolinetz  yes: https://github.com/fabric8io/openshift-elasticsearch-plugin/pull/134/files#diff-31ab0e029350fabbb48cd9560bb451f4R138

src/main/java/io/fabric8/elasticsearch/plugin/acl/ACLDocumentManager.java
```diff
@@ -0,0 +1,287 @@
...
+    class ExpireOperation implements ACLDocumentOperation {
+
+        private long now;
+
+        public ExpireOperation(long currentTimeMillis) {
+            this.now = currentTimeMillis;
+        }
+
+        @Override
+        public void syncDocuments(Collection<SearchGuardACLDocument> docs) {
```
richm: this method name is misleading - is it really `expireDocuments`

src/main/java/io/fabric8/elasticsearch/plugin/acl/ACLDocumentManager.java
```diff
@@ -0,0 +1,287 @@
...
+        }
+
+        @Override
+        public BulkRequest buildRequest(Client client, BulkRequestBuilder builder, Collection<SearchGuardACLDocument> docs) throws IOException{
+
+            for (SearchGuardACLDocument doc : docs) {
+                UpdateRequestBuilder update = client
+                        .prepareUpdate(searchGuardIndex, doc.getType(), SEARCHGUARD_CONFIG_ID)
+                        .setDoc(doc.toXContentBuilder());
+                if(doc.getVersion() != null) {
```
richm: under what circumstances would there not be a document version here?  IIUC, the document version is the key to ensuring the atomicity of updates?

jcantrill: The only case I can think of is when its a 'create' operation which would mean initialization.  It also could be an artifact of testing but I am unable to concretely say.

src/main/java/io/fabric8/elasticsearch/plugin/acl/ACLDocumentManager.java
jcantrill: It's an impl of an interface.  I will rename to make it more generic

jcantrill:
@ewolinetz @richm based on [1] and manual testing of this code with 2 ES nodes, believe the initial blockers are resolved.  Thoughts on merging?  

[1] https://github.com/openshift/origin-aggregated-logging/pull/1177

ewolinetz:
@jcantrill can we get the license check and the test to pass?

jcantrill:
@ewolinetz @richm @lukas-vlcek  any concerns moving forward with merging this PR

jcantrill:
[(3) Implement Single Searchguard Index [epic-ois-es5]](https://trello.com/c/d3Fadd6D/557-3-implement-single-searchguard-index-epic-ois-es5)

ewolinetz:
LGTM

src/main/java/io/fabric8/elasticsearch/plugin/ConfigurationSettings.java
```diff
@@ -91,7 +90,9 @@
      * Kibana index.
      */
     static final String OPENSHIFT_DYNAMIC_ENABLED_FLAG = "openshift.acl.dynamic.enabled";
-    
+
+    static final String OPENSHIFT_ACL_EXPIRE_IN_MILLIS = "openshift.acl.expire_in_millis";
```
ewolinetz: Can we also add a README entry for this?

jcantrill:
[merge]

jcantrill:
[merge]

PR #141, opened 2018-06-28
fix 1596099. Update permissions with IndexRequest instead of UpdateRe…

jcantrill:
…quest to get overwrite behavior

This PR fixes https://bugzilla.redhat.com/show_bug.cgi?id=1596099 by writing ACL's using IndexRequest instead of UpdateRequest which relies on optimistic concurrency and document merge behavior

### DIFF
--- a/src/test/resources/io/fabric8/elasticsearch/plugin/searchguard_roles_acl.yml
+++ b/src/test/resources/io/fabric8/elasticsearch/plugin/searchguard_roles_acl.yml
@@ -13,7 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
+sg_unrecognized_child:
+  foo: [bar]
+  
 sg_role_kibana:
   cluster:
       - cluster:monitor/nodes/info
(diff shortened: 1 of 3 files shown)


src/test/resources/io/fabric8/elasticsearch/plugin/searchguard_roles_acl.yml
```diff
@@ -13,7 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
+sg_unrecognized_child:
```
ewolinetz: how is this test entry used in our tests?

ewolinetz:
question about the test file entry, otherwise LGTM

src/test/resources/io/fabric8/elasticsearch/plugin/searchguard_roles_acl.yml
jcantrill: Tests a NullPointer I discovered while fixing this issue https://github.com/fabric8io/openshift-elasticsearch-plugin/blob/2.4.4/src/test/java/io/fabric8/elasticsearch/plugin/acl/SearchGuardRoleACLTest.java#L158

jcantrill:
[merge]

PR #150, opened 2018-07-20
fix some strange issue using the fowarded access token

jcantrill:
Found an issue where the access token was being skipped even though it was properly provided by the new proxy

### DIFF
--- a/src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java
+++ b/src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java
@@ -56,7 +56,7 @@ import io.netty.channel.Channel;
 public class RequestUtils implements ConfigurationSettings  {
     
     private static final Logger LOGGER = Loggers.getLogger(RequestUtils.class);
-    public static final String AUTHORIZATION_HEADER = "Authorization";
+    public static final String AUTHORIZATION_HEADER = "authorization";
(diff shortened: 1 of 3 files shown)


src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java
```diff
@@ -78,16 +78,16 @@ public String getUser(RestRequest request) {
     }
     
     public String getBearerToken(RestRequest request) {
-        String token = "";
-        if (request.header(AUTHORIZATION_HEADER) != null) {
-            final String[] auth = StringUtils.defaultIfEmpty(request.header(AUTHORIZATION_HEADER), "").split(" ");
-            if (auth.length >= 2 && "Bearer".equals(auth[0])) {
-                token = auth[1];
+        String token = request.header(X_FORWARDED_ACCESS_TOKEN);
```
richm: does `request.header` do case insensitive comparison?  If not, do you need to somehow make all of the request headers lower case before checking?  Same with `request.header(AUTHORIZATION_HEADER)` below.

jcantrill: it is case insensitive to the value per the documentation.

jcantrill:
[merge]

PR #151, opened 2018-07-24
bump f8 client. sync w/ 2.4.4

jcantrill:
This PR:
* removes the need to specify a username header as was made in 2.4.4 line
* bumps the fabric8 kubernetes-client to latest uber jar to avoid jarhell
* crafts SARs manually to avoid deserialization permission issues when used with Elasticsearch
* preps code to re-enable the integration tests which were blocked by the client impl


src/main/java/io/fabric8/elasticsearch/plugin/OpenshiftAPIService.java
```diff
@@ -0,0 +1,139 @@
...
+    private static final String ACCEPT = "Accept";
+    private static final String CONTENT_TYPE = "Content-Type";
+    private static final String APPLICATION_JSON = "application/json";
+    private static final Logger LOGGER = Loggers.getLogger(OpenshiftAPIService.class);
+    
+    public String userName(final String token) {
+        Response response = null;
+        try (DefaultOpenShiftClient client = buildClient(token)) {
+            Request okRequest = new Request.Builder()
+                    .url(client.getMasterUrl() + "apis/user.openshift.io/v1/users/~")
```
richm: Does the `~` here mean "the user whose token this is"?

richm: Is there a way to request only the `metadata` or `metadata.name`?  Purely in the interest of reducing the response payload.

src/main/java/io/fabric8/elasticsearch/plugin/OpenshiftAPIService.java
```diff
@@ -0,0 +1,139 @@
...
+        try (DefaultOpenShiftClient client = buildClient(token)) {
+            Request request = new Request.Builder()
+                .url(client.getMasterUrl() + "apis/project.openshift.io/v1/projects")
+                .header(ACCEPT, APPLICATION_JSON)
+                .build();
+            Response response = client.getHttpClient().newCall(request).execute();
+            if(response.code() != RestStatus.OK.getStatus()) {
+                throw new ElasticsearchSecurityException("Unable to retrieve users's project list", RestStatus.fromCode(response.code()));
+            }
+            return new HashSet<String>(JsonPath.read(response.body().byteStream(),"$.items[*].metadata.name"));
```
richm: Is there a way to request only the `metadata` or `metadata.name`?  Purely in the interest of reducing the response payload.

src/main/java/io/fabric8/elasticsearch/plugin/OpenshiftAPIService.java
jcantrill: Correct.  This means 'give me info about the user who is associated with this token'.  I know of no way to reduce the size of the payload.  The oc binary does the same; it gets the entire JSON but only presents you with a formatted output

src/main/java/io/fabric8/elasticsearch/plugin/OpenshiftRequestContextFactory.java
```diff
@@ -90,9 +84,6 @@ public OpenshiftRequestContext create(final RestRequest request) throws Exceptio
         boolean isClusterAdmin = false;
         String user = utils.getUser(request);
         String token = utils.getBearerToken(request);
-        if(utils.hasUserHeader(request) && StringUtils.isBlank(token)) {
```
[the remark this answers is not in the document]
jcantrill: This is the same logic we have in v2.4.4.x.  We no longer need the user header because we no longer effectively need to trust the proxy or an incoming curl request.  We make the assumption if you have a valid token then we will determine who you actually are and allow the request to continue under those findings.  We took it out in the other version because there is no need to validate your real username against the one you say it is if we are always going to use the real one.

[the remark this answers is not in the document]
richm: /lgtm

jcantrill:
[merge]

PR #154, opened 2018-08-01
rely on official SG jar

jcantrill:
This PR pulls in the SG version instead of using the one we build

I'll have to fix the CI job to not install the dependent jar for the test to pass


jcantrill:
[test]

jcantrill:
[merge]

PR #156, opened 2018-08-07
fix #155. Fix permissions for KibanaIndexMode

### LINKED ISSUE #117: Ops user in kibana mode of unique does not have a rolemapping to their unique kibana index

### LINKED ISSUE #155: Kibana Index mode not properly reflected in permissions
We advertise permissions for the kibana index mode as:

```
    /**
     * The setting that determines the kibana index is used by users.  Valid values are one of the following:
     * 
     *   * unique (Default) - Each user gets a unique index for kibana visualizations (e.g. .kibana.USER_UUID)
     *   * shared_ops       - Users who are in an ops role will share an index (e.g. kibana_ops) while non ops users will 
     *                        have a unique index (e.g. .kibana.USER_UUID)
     *   * shared_non_ops   - Users who are in an ops role will share an index (e.g. kibana) while non ops users will 
     *                        share the default index (e.g. .kibana_non_ops)                      
     */
```
but we currently do not generate those permission that reflect this feature.  Following is an example for and operations user while set to unique. I would expect a role like 'gen_kibana_SOMEHASH':
```
  "gen_project_operations" : {
    "cluster" : [
      "CLUSTER_OPERATIONS"
    ],
    "expires" : "1533221660151",
    "indices" : {
      "*?*?*" : {
        "*" : [
          "INDEX_ANY_OPERATIONS"
        ]
      },
      "?operations?" : {
        "*" : [
          "INDEX_OPERATIONS"
        ]
      }
    }
  },
  "sg_role_prometheus" : {
    "cluster" : [
      "METRICS"
    ]
  },
  "sg_role_curator" : {
    "cluster" : [
      "CLUSTER_MONITOR"
    ],
    "indices" : {
      "*" : {
        "*" : [
          "READ",
          "MANAGE"
        ]
      }
    }
  },
  "sg_role_kibana" : {
    "cluster" : [
      "CLUSTER_COMPOSITE_OPS",
      "CLUSTER_MONITOR"
    ],
    "indices" : {
      "?kibana" : {
        "*" : [
          "INDICES_ALL"
        ]
      }
    }
  },
  "sg_role_fluentd" : {
    "cluster" : [
      "CLUSTER_MONITOR",
      "indices:data/write/bulk"
    ],
    "indices" : {
      "*" : {
        "*" : [
          "CRUD",
          "CREATE_INDEX"
        ]
      }
    }
  },
  "sg_project_operations" : {
    "indices" : {
      "*?*?*" : {
        "*" : [
          "READ",
          "indices:admin/validate/query*",
          "indices:admin/get*",
          "indices:admin/mappings/fields/get*"
        ]
      },
      "?operations?*" : {
        "*" : [
          "READ",
          "indices:admin/validate/query*",
          "indices:admin/get*",
          "indices:admin/mappings/fields/get*"
        ]
      }
    }
  },
  "sg_role_admin" : {
    "cluster" : [
      "CLUSTER_ALL"
    ],
    "indices" : {
      "*" : {
        "*" : [
          "ALL"
        ]
      }
    }
  },
  "gen_ocp_kibana_shared" : {
    "cluster" : [
      "CLUSTER_MONITOR_KIBANA"
    ],
    "expires" : "1533221660151",
    "indices" : {
      "*" : {
        "*" : [
          "INDEX_ANY_KIBANA"
        ]
      }
    }
  }
}
```

jcantrill:
This PR fixes #155  and will most likely need additional changes in sg_action_groups for permissions

Also fixes #117


jcantrill:
@ewolinetz @richm Please when looking at these changes consider looking at files like which is my expectation of what the perm file(s) should be according to how we define KibanaIndexMode.

[1] https://github.com/fabric8io/openshift-elasticsearch-plugin/blob/2abc1531a05dd5e067d4b57f8ced4dca164a5192/src/test/resources/io/fabric8/elasticsearch/plugin/project_strategy_roles_shared_non_ops_kibana_mode.yml

src/main/java/io/fabric8/elasticsearch/plugin/acl/BaseRolesSyncStrategy.java
```diff
@@ -19,14 +19,15 @@
...
+import io.fabric8.elasticsearch.plugin.KibanaIndexMode;
 import io.fabric8.elasticsearch.plugin.OpenshiftRequestContextFactory;
 import io.fabric8.elasticsearch.plugin.OpenshiftRequestContextFactory.OpenshiftRequestContext;
 
 public abstract class BaseRolesSyncStrategy implements RolesSyncStrategy {
 
     protected SearchGuardRoles roles;
     private final String userProfilePrefix;
-   
+
```
richm: extraneous space

src/main/java/io/fabric8/elasticsearch/plugin/acl/UserRolesSyncStrategy.java
```diff
@@ -33,40 +33,32 @@ public UserRolesSyncStrategy(SearchGuardRoles roles, String userProfilePrefix, S
         this.kibanaIndexMode = kibanaIndexMode;
         this.expire = String.valueOf(expiresInMillis);
     }
-
+
```
richm: extra space

src/main/java/io/fabric8/elasticsearch/plugin/acl/BaseRolesSyncStrategy.java
```diff
@@ -49,10 +50,20 @@ protected String formatKibanaIndexName(OpenshiftRequestContext context, String k
...
     
-    protected String formatKibanaRoleName(OpenshiftRequestContext context) {
-        if (context.isOperationsUser()) {
-            return SearchGuardRolesMapping.KIBANA_SHARED_ROLE;
-        } else {
+    public static String formatKibanaRoleName(OpenshiftRequestContext context) {
+        final String kibanaIndexMode = context.getKibanaIndexMode();
+        if (context.isOperationsUser() ) {
+            switch (kibanaIndexMode) {
+            case KibanaIndexMode.UNIQUE:
```
ewolinetz: can we indent our `case`s? IMO it makes switch statements easier to read...

src/main/java/io/fabric8/elasticsearch/plugin/acl/UserRolesMappingSyncStrategy.java
```diff
@@ -39,15 +39,15 @@ public UserRolesMappingSyncStrategy(SearchGuardRolesMapping mapping, long expire
...
+        String kibanaRoleName = BaseRolesSyncStrategy.formatKibanaRoleName(context);
+        builder.addUser(kibanaRoleName, user)
+            .expire(getExpires());
         if (context.isOperationsUser()) {
-            builder.addUser(SearchGuardRolesMapping.ADMIN_ROLE, user);
-            builder.addUser(SearchGuardRolesMapping.KIBANA_SHARED_ROLE, user);
-            builder.expire(getExpires());
+            builder.addUser(SearchGuardRolesMapping.ADMIN_ROLE, user)
+                .expire(getExpires());
         } else {
```
ewolinetz: is it correct to have an else here? wouldn't' we want an admin to pick up the role in addition to the admin_role? we do something similar in `ProjectRolesMappingSyncStrategy.java`

src/main/java/io/fabric8/elasticsearch/plugin/acl/UserRolesSyncStrategy.java
```diff
@@ -33,40 +33,32 @@ public UserRolesSyncStrategy(SearchGuardRoles roles, String userProfilePrefix, S
...
+                .setActions(ALL, ALL, KIBANA_ROLE_ALL_INDEX_ACTIONS)
+                .expires(expire);
+        builder.addRole(kibRole.build());
+
+
         //think this can be statically added to roles doc
         if (context.isOperationsUser()) {
             RoleBuilder opsRole = new RoleBuilder(SearchGuardRolesMapping.ADMIN_ROLE)
                     .setClusters(OPERATIONS_ROLE_CLUSTER_ACTIONS)
                     .setActions("?operations?", ALL, OPERATIONS_ROLE_OPERATIONS_ACTIONS)
```
ewolinetz: is this action still required if we have `ALL`?

src/main/java/io/fabric8/elasticsearch/plugin/acl/BaseRolesSyncStrategy.java
jcantrill: I believe this is simply a space between the member  variables and the constructor

src/main/java/io/fabric8/elasticsearch/plugin/acl/BaseRolesSyncStrategy.java
jcantrill: We cannot based on the current style rule.  It will have to be changed if that's what we want.  I, unfortunately, don't know what the quick win is to make that happen

src/main/java/io/fabric8/elasticsearch/plugin/acl/UserRolesMappingSyncStrategy.java
jcantrill: The code will setup the kibana role as well as access to the projects they can see.  In this case, admin users will always be able to see the entire set of indices.

src/main/java/io/fabric8/elasticsearch/plugin/acl/UserRolesSyncStrategy.java
jcantrill: Agree.  We can remove. I wasn't certain why we ever made this division.  The only possibility is if we desire to distinguish actions from one set of indices vs another set but I presume `ALL` would always be the winner

jcantrill:
@richm @ewolinetz Updated per CR

jcantrill:
@richm @ewolinetz I've updated the code to match the KibanaIndexMode as advertised as well as fix to pass https://github.com/openshift/origin-aggregated-logging/pull/1273.

**NOTE:**  Added several integration tests to match those we have in the logging repo.

richm:
openshift/origin-aggregated-logging#1273 passed so I'm assuming this PR is good to go
/lgtm

jcantrill:
[merge]

PR #157, opened 2018-09-04
bug 1622822. Add uuid to acl docs

jcantrill:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1622822
* adds uid to permissions
* adds uid to index-pattern


jcantrill:
yes.  all ACLs are supposed to expire every 60s

jcantrill:
@ewolinetz @richm updated to include modifications to index pattern to include uid

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaUtils.java
```diff
@@ -0,0 +1,110 @@
...
+            return Project.EMPTY;
+        }
+
+        Matcher matcher = reIndexPattern.matcher(index);
+        if (matcher.matches()) {
+            return new Project(matcher.group("name"), matcher.group("uid"));
+        }
+        return new Project(index, null);
+    }
+
```
richm: if the index is `project.*`, then index = `project.*` and uid = `null`
then `formatIndexPattern` will return `project.project.*.*`?

src/test/resources/io/fabric8/elasticsearch/plugin/project_strategy_roles_shared_non_ops_kibana_mode.yml
```diff
@@ -11,9 +11,9 @@ gen_ocp_kibana_shared_non_ops:
 gen_project_foo_bar:
   expires: '10'
   indices:
-    ?project?foo?bar?*:
+    ?project?foo?bar?123abc?*:
```
richm: in this case - `project` is the project prefix, `foo` is the namespace name?  what is `bar`?  is `123abc` the uuid?

src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaUtils.java
richm: ok - hard to tell from the flow - but if you are sure that this method is never called with "project.*" then that's good enough for me

jcantrill:
[merge]

PR #158, opened 2018-09-25
enhancement #LOG192. Impl Openshift Token Auth

jcantrill:
Enhancement to add bearer token auth https://jira.coreos.com/browse/LOG-192

This PR additionally allows us to add static groups which should also satisfy the prometheus metrics usecase and removing the proxy for that use.

### DIFF
--- a/src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
+++ b/src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
@@ -16,27 +16,20 @@
 
 package io.fabric8.elasticsearch.plugin.acl;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map.Entry;
 import java.util.function.UnaryOperator;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.logging.log4j.Logger;
-import org.elasticsearch.ElasticsearchSecurityException;
 import org.elasticsearch.client.Client;
 import org.elasticsearch.client.node.NodeClient;
 import org.elasticsearch.common.logging.Loggers;
-import org.elasticsearch.common.util.concurrent.ThreadContext.StoredContext;
-import org.elasticsearch.rest.BytesRestResponse;
+import org.elasticsearch.common.util.concurrent.ThreadContext;
 import org.elasticsearch.rest.RestChannel;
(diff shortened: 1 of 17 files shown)


src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
```diff
@@ -82,77 +66,52 @@ public DynamicACLFilter(final PluginSettings settings,
...
-                    requestContext = contextFactory.create(request);
-                    request = utils.modifyRequest(request, requestContext, channel);
-                    if (requestContext != OpenshiftRequestContext.EMPTY) {
-                        logRequest(request);
-                        final String kbnVersion = getKibanaVersion(request);
-                        kibanaSeed.setDashboards(requestContext, client, kbnVersion, cdmProjectPrefix);
-                        aclManager.syncAcl(requestContext);
-                    }
-
+            if (enabled && threadContext.getTransient(OPENSHIFT_REQUEST_CONTEXT) != null) {
```
ewolinetz: side question, but do we really need to check for `enabled` anymore? should we just deprecate that feature?

ewolinetz:
Would this PR also let us remove the oauth proxy container?

src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
jcantrill: I have wondered that for a while now and fully onboard with removing this all together.  I'm not sure how or why we would enable the feature.  It would be better if it were possible to hot reload it and maybe we would see an advantage for diagnosis or development when things are disabled.  I'll push changes to remove it.  maybe in a separate pr

jcantrill:
@ewolinetz I went ahead and removed the 'enabled' configs.

ewolinetz:
@jcantrill can you add to the README that we removed the flags to enable and disable functionality?
Otherwise lgtm

ewolinetz:
i dont think the bot is listening here...

richm:
yeah - I think Jeff will merge it manually if we approve - so just Request Changes for now

jcantrill:
Merging as https://github.com/openshift/origin-aggregated-logging/pull/1363 finally passed its tests

PR #159, opened 2018-09-28
Fix delete index pattern v2

### LINKED ISSUE #135: Deleting an index-pattern in Kibana causes a 409 response
Deleting an index-pattern in Kibana causes it to blowup and a 409 is returned for at least v5.6.9

jcantrill:
The PR:
* fixes #135 
* Moves response rewrite for Kibana index to the rest level

Notes:
* It wasn't readily obvious to me the plugin would creating the index-pattern we deleted if it was one that we seed on request
* If the response includes Kibana indices they will be reset to the default.  I'm fairly certain this is not a regression but the behavior we saw prior
* need to rebase on #158


src/main/java/io/fabric8/elasticsearch/plugin/acl/DynamicACLFilter.java
```diff
@@ -33,6 +33,7 @@
 import io.fabric8.elasticsearch.plugin.OpenshiftRequestContextFactory.OpenshiftRequestContext;
 import io.fabric8.elasticsearch.plugin.PluginSettings;
 import io.fabric8.elasticsearch.plugin.kibana.KibanaSeed;
+import io.fabric8.elasticsearch.plugin.rest.RestChannelInterceptor;
```
richm: Is this api available in ES 2.x e.g. can we backport this to earlier releases?

jcantrill: This was code I inserted and may be something we could backport but do we have need? 3.11 is considered our long term release which is based on 5.x.  We should refrain from backporting unless we have a good reason (e.g. bug, etc)

ewolinetz:
Do we have tests for this? I'd like to make sure that it can still correctly return the different types of responses to Kibana without it breaking due to us tampering with the query/response.

jcantrill:
@ewolinetz added tests.  GetFieldMappings was not possible as it has no public constructor

jcantrill:
[merge]

PR #161, opened 2018-10-24
LOG-226. Allow SAR of nonResourcedURL

jcantrill:
This PR fixes https://jira.coreos.com/browse/LOG-226 

which allows using a nonResourceURL to check SARs.  This will allow us to rely on the clusterrole provided by the cluster-monitoring-operator


jcantrill:
[merge]

PR #166, opened 2018-12-05
fix upgrade of kibana

jcantrill:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1652224 by recreating the 'latest' config object if needed


pom.xml
```diff
@@ -252,7 +250,10 @@
                 <artifactId>maven-failsafe-plugin</artifactId>
                 <version>2.20</version>
                 <configuration>
-                    <argLine>-da -Djava.security.manager -Djava.security.policy=${basedir}/src/test/resources/java.policy</argLine>
+                    <argLine>-da -Djava.security.manager
+                        -Djava.security.policy=${basedir}/src/test/resources/java.policy
+                        -Djdk.tls.rejectClientInitiatedRenegotiation=true
```
richm: ?

jcantrill: To remove the warning in the integration logs

src/it/java/io/fabric8/elasticsearch/KibanaIndexModeSharedOpsIntegrationTest.java
```diff
@@ -37,12 +38,12 @@ public void testNonAdminUserRetrievingDocuments() throws Exception {
         final String userName = "nonadminuser";
         //artificially seed user's kibanaIndex
         String kibanaIndex = getKibanaIndex(KibanaIndexMode.SHARED_OPS, userName, false);
-        givenDocumentIsIndexed(kibanaIndex, "config", "0", "myKibanaIndex");
+        givenDocumentIsIndexed(kibanaIndex, "config", "4.6.4", "myKibanaIndex");
```
ewolinetz: instead of hardcoding `4.6.4` everywhere can we move it into a test env var?

jcantrill: you mean like an actual environment variable or just a constant?  As long as what we seed is semantically older then current it wont make a difference

PR #170, opened 2019-04-01
bug 1694884. Default kibana index if blank

jcantrill:
Fix https://bugzilla.redhat.com/show_bug.cgi?id=1694884

### DIFF
--- a/src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
+++ b/src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
@@ -23,6 +23,7 @@ import java.util.Set;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.logging.log4j.Logger;
+import org.elasticsearch.action.get.GetResponse;
 import org.elasticsearch.common.collect.Tuple;
 import org.elasticsearch.common.logging.Loggers;
 import org.elasticsearch.common.settings.Settings;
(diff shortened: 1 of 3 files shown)


src/main/java/io/fabric8/elasticsearch/plugin/kibana/KibanaSeed.java
```diff
@@ -78,9 +79,10 @@ public void setDashboards(final OpenshiftRequestContext context, String kibanaVe
 
         if (action.v2() != null && !Project.EMPTY.equals(action.v2())) {
             boolean defaultIndexPatternExists = pluginClient.documentExists(context.getKibanaIndex(), INDICIES_TYPE, action.v2().getName());
-            boolean kibanaConfigExists = pluginClient.documentExists(context.getKibanaIndex(), CONFIG_DOC_TYPE, kibanaVersion);
-            if(!defaultIndexPatternExists || !kibanaConfigExists){
+            GetResponse config = pluginClient.getDocument(context.getKibanaIndex(), CONFIG_DOC_TYPE, kibanaVersion);
+            if(!defaultIndexPatternExists || !config.isExists() || StringUtils.isBlank(kibanaUtils.getDefaultIndexPattern(config))){
```
richm: can `pluginClient.getDocument` return `null`?

jcantrill: It's a wrapper around the ES calls so I don't see how.  If response doesnt find anything, AFAIK it returns an object because it has the exist method.  If you prefer, I could update, for safety to additionally check null'

richm: If it can't be `null`, then no need to check

jcantrill:
[merge]

PR #171, opened 2019-04-04
bug 1694884. Don't explicitly define kibana index replica

jcantrill:
This PR fixes https://bugzilla.redhat.com/show_bug.cgi?id=1667801 by removing the explicit replica setting when creating the index and let it rely on the template


jcantrill:
[merge]

PR #176, opened 2019-04-25
collate configuration items

Gallardot:
If we want to use some configuration items, we should declare them on the OpenShiftElasticSearchPlugin.

I did some proof-reading.


jcantrill:
[merge]

PR #186, opened 2019-08-30
fix missing token for OpenShiftAPIService requests

jcantrill:
This PR explicitly sets a user's token on the request as a header when making calls to the Openshift API server to resolve an issue where there was unexpected response when executing a SAR

### DIFF
--- a/src/main/java/io/fabric8/elasticsearch/plugin/OpenshiftAPIService.java
+++ b/src/main/java/io/fabric8/elasticsearch/plugin/OpenshiftAPIService.java
@@ -65,6 +65,7 @@ public class OpenshiftAPIService {
         try (DefaultOpenShiftClient client = factory.buildClient(token)) {
             Request okRequest = new Request.Builder()
                     .url(client.getMasterUrl() + "apis/user.openshift.io/v1/users/~")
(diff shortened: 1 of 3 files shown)


jcantrill:
@richm validating fix via https://github.com/openshift/origin-aggregated-logging/pull/1735

src/main/java/io/fabric8/elasticsearch/plugin/OpenshiftAPIService.java
```diff
@@ -65,6 +65,7 @@ public String userName(final String token) {
         try (DefaultOpenShiftClient client = factory.buildClient(token)) {
```
richm: `factory.buildClient(token)` - what does `buildClient` do with the `token`?

jcantrill: It was my understanding that it added the to the header, but we are using the raw client this one wrappers so not sure about what it does.  The reality is we could use a straight REST client without kubernetes-client as we don't use any of the features it provides other then maybe the CA look up, but Its likely not worth the effort to remove it.

jcantrill:
[merge]

PR #188, opened 2019-09-16
Bug 1746377: Fetch indices instead of search to seed index patterns

jcantrill:
This PR fixes https://bugzilla.redhat.com/show_bug.cgi?id=1746377 by:

* retrieving and comparing against a list of indices instead of using a search result


jcantrill:
[merge]

PR #190, opened 2020-07-01
Bump version of fabric8.io Kubernetes Client

csrwng:
Needed to include the fix in https://github.com/fabric8io/kubernetes-client/pull/2227


csrwng:
@jcantrill ptal - I am using the earliest version of the k8s client that includes the desired fix. Is that ok? or should it be the most recent version?

jcantrill:
> @jcantrill ptal - I am using the earliest version of the k8s client that includes the desired fix. Is that ok? or should it be the most recent version?

This should be fine.

csrwng:
Updated the PR to directly include jackson libraries overriding the version imported by both elasticsearch and the kube client.

PR #191, opened 2020-07-01
Disable HTTP2 in communication with OpenShift

csrwng:
Disables HTTP/2 protocol when communicating with OpenShift API Server
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1835396


csrwng:
@jcantrill ptal ... this is an alternative to updating the k8s client dependency.

csrwng:
/close
This doesn't fix the issue

PR #192, opened 2020-07-02
Disable HTTP2 in communication with OpenShift

csrwng:
This fixes https://bugzilla.redhat.com/show_bug.cgi?id=1835396


csrwng:
@jcantrill ptal - I have now tested this fix and it does work. It doesn't make any changes to dependencies so I would think it shouldn't give you trouble about conflicting dependencies.

